PR #31, opened 2016-09-27
Removes passing of parameters that are already in scope

JPrevost:


JPrevost:
FYI @szendeh @hakbailey @matt-bernhardt @carlj11 this addresses the inconsistent passing  of parameters vs using the params scope @szendeh pointed out in our discussion.

Issue #25, opened 2016-09-22
Fancy It Up


JPrevost:

JPrevost:
We'll handle this in Jira (with tickets that are likely more helpful).

PR #39, opened 2016-09-28
Adds 'more search tools' below results

matt-bernhardt:


app/views/search/bento.html.erb
```diff
@@ -1,29 +1,46 @@
...
+  <div class="col-sm-4">
+    <h2>Website and Guides</h2>
+    <div id="website_content">
+      Loading results...
+      <i class="fa fa-spinner fa-spin"></i>
+    </div>
   </div>
 </div>
 
+<div class="row">
```
JPrevost: Totally minor at this stage, but can you add some visual separation between the results and this message?

An ```<hr />``` or something simple would be enough for now.

app/views/search/bento.html.erb
```diff
@@ -1,29 +1,46 @@
...
+  <h2>Not finding what you're looking for? Try these other tools:</h2>
+  <dl>
+    <dt><a href="//libraries.mit.edu/vera">Vera</a></dt>
+    <dd>Best for e-journals and databases owned or licensed by the MIT Libraries. Don’t search for individual articles here.</dd>
+    <dt><a href="//libraries.mit.edu/worldcat">WorldCat</a></dt>
+    <dd>Best for books and other items not owned by the MIT Libraries. Using this interface, you can request items from <a href="//libraries.mit.edu/borrow-direct">Borrow Direct</a> or <a href="//libraries.mit.edu/ilb">Interlibrary Borrowing (ILB)</a>.</dd>
+  </dl>
+  <h3>Specialized search tools &amp; guides</h2>
```
JPrevost: This proposed section is likely to be contentious. Anything by architecture is fine with me ;)

matt-bernhardt: Yeah, I grabbed a section of content from libraries.mit.edu/search as a conversation starter - I'm curious what the response will be.

app/views/search/bento.html.erb
matt-bernhardt: Yeah, the spacing is pretty close. I wasn't sure whether just adding pixels would be enough (hopefully) or if we really would need a line. I'm generally against "draw a line" as a separation strategy.

I'll keep playing.

JPrevost: Mostly we just need to make it showable and hope @frrrances eventually stops screwing around with other projects and starts working on #25 ;)

PR #41, opened 2016-10-03
Adds Google Analytics

### LINKED ISSUE #26: Add Google Analytics
We can get fancier later, but for now Google Analytics is good enough.

matt-bernhardt:
This takes the stock GA code, de-minimizes it to pass the eslint check on CodeClimate, and adds a ternary check of the domain name to swap between registering activity to the test or production account.

As a first step, I think the goal is just to track pages. A further goal would be to adopt event tracking in order to record the items returned from the various tools, or to record where the user moves from the result page.

A second needed change before this launches is probably to re-consider how we check for the app being in production. I'd welcome @JPrevost thoughts here.

(I'm also happy to squash this branch before merging - so far I've been leaving the commits separate)

When merged, this closes #26.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,3 +30,4 @@ The query will be appended.
 ## Optional Environment Variables
 
 - `RESULTS_PER_BOX`: defaults to 3
+- `GOOGLE_ANALYTICS`: Google Analytics property ID for monitoring application use
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -21,6 +21,28 @@
     <%= favicon_link_tag 'favicon.ico', :rel => 'shortcut icon' %>
 
     <%= javascript_include_tag "application" %>
+    <% if (ENV['GOOGLE_ANALYTICS']) %>
+    <script type="text/javascript">
+      // Google Analytics
+
+      (function(i,s,o,g,r,a,m){
+        i['GoogleAnalyticsObject']=r;
+        i[r]=i[r]||function(){
+          (i[r].q=i[r].q||[]).push(arguments)
+        };
+        i[r].l=1*new Date();
+        a=s.createElement(o);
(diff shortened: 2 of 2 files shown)


matt-bernhardt:
In addition to the above, the final implementation will also probably need some degree of event tracking. For now, this just records pageviews - which is exceedingly minimal.

JPrevost:
I need to think about how to not run against any GA during full stack tests. Or even in local development I often wouldn't want to send stuff to GA.

It shouldn't be a huge change. Mostly I think we want to check for an GA_PROPERTY in ENV and only run this if so (and use that GA_PROPERTY value instead of storing property values in the code).

To be able to read ENV you'll need to rename this to `static.js.erb` if you want to take a look at that before I have a chance to think more about it.

matt-bernhardt:
That makes sense. I'll rework this to check for an ENV variable and perform
this only if present. There will probably be times when we _do_ want to
send data to GA in testing, but I can see not wanting to do that in all
cases.

JPrevost:
@matt-bernhardt Sorry for misleading information previously. I think this is best done in the application template and not in an external stylesheet due to the way the application pipeline only compiles at deploy time for production applications. This would work fine in dev environments, but deployed it will not work as intended. I.E. changing the ENV setting (or removing it) will have no effect until a redeploy when the pipeline recompiles.

So... can you take your script and move it back to the `application.html.erb` as this PR started. Thanks!

matt-bernhardt:
I _think_ this may be ready for squashing now? The presence/absence of `GOOGLE_ANALYTICS` in the environment works on my local box, but I don't have access to add the key to Heroku.

JPrevost:
Yeah, squash and merge. I added / removed the key on Heroku and it worked as expected. I'll transfer to the shared account when I'm at stabler wifi.

PR #42, opened 2016-10-03
Adds Codeclimate-suggested configuration files

matt-bernhardt:
These are the files that CodeClimate is asking we add. So far I've just run the suggested patch to create these, so if there are better values in the Rails community then I'm happy to use those. Mostly, I just want to be able to run eslint locally on the `matt/analytics` branch while I try and figure out what refactoring is needed.

Tagging @JPrevost for review


JPrevost:
It's not clear to me why this is required. Can you help me understand when CodeClimate needed this?

matt-bernhardt:
This is what CodeClimate is reporting on its interface:
image
...which is linked to https://codeclimate.com/repos/57c6e4b874ebe82931000266/analysis_config . This PR applies the suggested patch. I've used these files on other projects to add/remove specific tests and otherwise control what CodeClimate does.

JPrevost:
Are you changing any default values or just including the defaults?

matt-bernhardt:
This just uses the default values for now.

JPrevost:
My preference would be to continue to use the auto generated defaults until we need to change something unless this was causing problems for you. This way we'd continue to get any upstream improvements as they'd just keep generating them for us on the fly. I'm not super committed to either path though so feel free to either merge or close based on how you feel after hearing my thoughts on it :)

matt-bernhardt:
Reading your thoughts, I'm fine with letting things ride until we need to change from the defaults. I'll set up a new PR if/when that occurs.

PR #45, opened 2016-10-05
[WIP] First stab at recording interaction events

### LINKED ISSUE #43: Use Event Tracking for better analytics
Now that #41 is merged, we can consider what sort of event tracking would help us better evaluate how users interact with this system. I anticipate either or both of two possibilities:

1. Record what gets displayed. This would involve a scan down the displayed titles, recording basic information. This would help characterize / improve what is actually being returned by the various APIs - for example issues like #22 or #23

2. Record what gets interacted with. Do users click on the title to see a single item? The "View all ###" link to just get to the tool? Run another search? Follow a link to some other tool?

This type of information is currently being recorded within BartonPlus, and I think it would be useful to also have in this context.

matt-bernhardt:
This sets up event tracking when the user clicks on `a` tags in one of four defined regions: the three bento results areas, and the "other tools" area. What regions we focus on are obviously negotiable, as are how we call the listener in the first place.

If you look at the test discovery profile within GA, you'll see what events are being received.

I'd like to hear from @JPrevost and @szendeh about the technical side of this. The decision about whether to track this, or what to record at all, is probably something that should happen in Basecamp.

When merged, this would close #43

### DIFF
--- a/app/assets/javascripts/static.js
+++ b/app/assets/javascripts/static.js
@@ -1,2 +1,14 @@
 // Place all the behaviors and hooks related to the matching controller here.
 // All this logic will automatically be available in application.js.
+
+function TrackLinks( element ) {
+  $( element ).click( function( e ) {
+    fields = {
+      hitType: 'event',
+      eventCategory: 'Bento',
+      eventAction: $( this ).parents( '.region' ).data( 'region' ),
+      eventLabel: e.target.text
+    };
+    ga( 'send', fields );
+  });
+}
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -41,6 +41,12 @@
 
       ga('create', '<%= ENV['GOOGLE_ANALYTICS'] %>', 'auto');
       ga('send', 'pageview');
+
+      // Interaction tracking
+      $(document).ready(function(e) {
+        TrackLinks( $( '.region a' ) );
+      });
+
     </script>
     <% end %>
   </head>
--- a/app/views/search/bento.html.erb
+++ b/app/views/search/bento.html.erb
@@ -3,7 +3,7 @@
 <div class="row">
   <div class="col-sm-4">
     <h2>Articles and Stuff</h2>
-    <div id="articles_content">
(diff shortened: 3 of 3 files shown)


szendeh:
Hey so as we spoke about, one easy way to access the elements inside the content divs is to use

http://api.jquery.com/ajaxcomplete/

eg instead of:

$(document).ready(function() {
    xyz

you could do:

 $(document).ajaxComplete(function() {
    xyz

The 'complete' event is only triggered once the ajax call is complete *and, in the case of load, the html and the dom are updated*.

So that means your code could add listeners to specific elements you expect to be there, instead of on the entire div.

Also, you could directly call the function you that adds the listeners by adding a function call to the end of the load function call eg instead of:

$('#articles_content').load("/search/search?q=<%= URI.encode(params[:q]) %>&target=articles");

$('#articles_content').load("/search/search?q=<%= URI.encode(params[:q]) %>&target=articles", yourFunction);

You'd have to add it to each of the load calls, though.

I hope that helps, I'm not sure what you should be tracking, that's something we'll need to think about and maybe ask UX or something.

matt-bernhardt:
Thanks for these options - I tried the `ajaxComplete` method, but made more headway with just chaining the function at the end of the loading function. Yes, this is duplicative, but no more so than it was before. We could probably refactor this relatively easily, but for now this gets the job done.

Thoughts, @JPrevost ?

JPrevost:
We can refactor later to reduce code duplication so I have no issues with that aspect.

👍

This following is not a blocking comment, so you can squash and merge or consider making some minor changes:

It is unclear to me what the preference for spaces around parameters or single / double quotes is in the JavaScript world... and I don't have strong feelings about how to do things in JavaScript, but I think we've probably got too many inconsistencies in this single line (for example):

`eventAction: $( this ).parents(".region").data('region'),`

(spaces around params, no spaces around params, single quotes, double quotes).

matt-bernhardt:
Good eye - I've squashed and gone through these changes and tried to use quotes/etc more consistently. Once those come back clean, I'll merge.

PR #47, opened 2016-10-11
Adds Worldcat search panel

matt-bernhardt:


app/models/normalize_worldcat.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  private
+
+  # Extract the information we care about from the raw results and add them
+  # return them as an array of {result}s
+  def extract_results(xml, norm)
+    xml.xpath('//xmlns:oclcdcs').each do |item|
+      result = Result.new(item.xpath('dc:title').text.strip,
+                          item.xpath('dc:date').text.strip,
+                          'http://www.worldcat.org/oclc/' + item.xpath('oclcterms:recordIdentifier[not(@xsi:type)]').text.strip,
```
JPrevost: I'd suggest extracting this link creating logic to it's own method to make this easier to read:

```
  def item_link(item)
    'http://www.worldcat.org/oclc/' +
      item.xpath('oclcterms:recordIdentifier[not(@xsi:type)]').text.strip
  end
```

PR #49, opened 2016-10-13
Truncate long author lists

### LINKED ISSUE #22: Do something about lots of authors
Some items have lots and lots of authors. It's unlikely displaying them all is useful and it sure looks bad.

matt-bernhardt:
This closes #22

### DIFF
--- a/app/models/result.rb
+++ b/app/models/result.rb
@@ -12,4 +12,10 @@ class Result
     @url = url
     @type = type
   end
+
+  # View-type method for returning a truncated list of authors.
+  def truncated_authors
+    return authors if authors.length <= ENV['MAX_AUTHORS'].to_i
+    authors[0...ENV['MAX_AUTHORS'].to_i] << 'et al'
+  end
 end
--- a/README.md
+++ b/README.md
@@ -28,6 +28,8 @@ The query will be appended.
 ex: `website,books,articles`
 - `GOOGLE_API_KEY`: your Google Custom Search API key
 - `GOOGLE_CUSTOM_SEARCH_ID`: your Google Custom Search engine ID
+- `MAX_AUTHORS`: the maximum number of authors displayed in any record.
+If exceeded, 'et al' will be appended after this number.
 
 ## Optional Environment Variables
 
--- a/app.json
+++ b/app.json
@@ -35,6 +35,9 @@
     "LANG": {
       "required": true
     },
+    "MAX_AUTHORS": {
+      "required": true
+    },
     "RACK_ENV": {
       "required": true
     },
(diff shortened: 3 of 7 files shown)


matt-bernhardt:
I'm still working through the most efficient way to call this. Right now it works for something like

```
foo = Result.new(...)
foo.authors = foo.truncate_authors(['foo','bar','baz','none'])
```

app/models/result.rb
```diff
@@ -12,4 +12,11 @@ def initialize(title, year, url, type)
     @url = url
     @type = type
   end
+
+  # Testing something
+  def truncate_authors(authors)
+    return @authors if !ENV['MAX_AUTHORS'] || authors.length <= ENV['MAX_AUTHORS'].to_i
```
JPrevost: Negative conditions are sometimes hard to grok. Rails allows `unless` to allow you to not have to `!` things.

Also, I think we can just make this a required ENV and not do that check at all.

matt-bernhardt:
Ok, cool. I'll make those changes and adjust the readme.

JPrevost:
When making use of new ENV, especially required ones like this, you'll want to update `app.json` to require it as well which tells the PR builds to copy that ENV setting from the Staging app.

test/models/normalize_eds_test.rb
```diff
@@ -56,6 +56,19 @@ class NormalizeEdsTest < ActiveSupport::TestCase
     end
   end
 
+  test 'can handle many authors' do
```
JPrevost: This test is not really needed as the code you are adding operates on Results and thus doesn't care about how the authors were originally added to a Result in the first place.

test/models/normalize_worldcat_test.rb
```diff
@@ -59,4 +59,16 @@ class NormalizeWorldcatTest < ActiveSupport::TestCase
       assert_equal(0, query['total'])
     end
   end
+
+  test 'searches with many authors get truncated list' do
```
JPrevost: Also not really necessary (see comment about test in `test/models/normalize_eds_test.rb`)

app/models/result.rb
```diff
@@ -12,4 +12,10 @@ def initialize(title, year, url, type)
     @url = url
     @type = type
   end
+
+  # View-type method for returning a truncated list of authors.
+  def truncated_authors
+    return authors if authors.length <= ENV['MAX_AUTHORS'].to_i
+    authors[0..ENV['MAX_AUTHORS'].to_i - 1] << 'et al'
```
JPrevost: A more ruby way to write this is:
`authors[0...ENV['MAX_AUTHORS'].to_i] << 'et al'`

See: http://ruby-doc.org/core-2.3.1/Range.html

PR #50, opened 2016-10-14
Allows boxes to be toggled

JPrevost:
A session variable persists the toggle.
A default is set in ENV.
Links below the footer trigger the toggles. If this feature persists beyond initial testing, we'll probably want to handle that differently but for now it seems fine.

There is also some DRY cleanup in the views moving some similar code to partials with local variables for the changing bits.


app/helpers/bento_helper.rb
```diff
@@ -0,0 +1,9 @@
+module BentoHelper
+  def box_enabled?(box)
+    session[:boxes].include?(box)
+  end
+
+  def box_width
+    (12 / session[:boxes].count).to_i
```
matt-bernhardt: I'll defer to @frrrances here, but I want to say this might be a candidate for using flex-box for layout - rather than having to calculate the number of panels in the display and set widths accordingly, we should be able to have widths auto-calculated.

This isn't to say I object to this approach, especially for now - just that this might need to be revisited when the "real" UI design gets implemented.

.rubocop.yml
```diff
@@ -17,3 +17,12 @@ ClassLength:
 
 Style/FrozenStringLiteralComment:
   Enabled: false
+
+Metrics/BlockLength:
+  CountComments: false  # count full line comments?
+  Max: 25
```
JPrevost: This is a new metric in rubocop and I'm mostly just telling it to ignore the normal stuff we ignore plus the `test_helper` which has a huge block that is really not worth rewriting at this time.

app/helpers/bento_helper.rb
JPrevost: Definitely. 100% of this UI is going to be rewritten so I haven't been worried to much about it.

app/views/search/_placeholders.html.erb
```diff
@@ -0,0 +1,7 @@
+<div class="col-sm-<%= box_width %>">
+  <h2><%= heading %></h2>
+  <div id="<%= id %>">
```
JPrevost: Ah, got it. I noticed the one you added for worldcat didn't have those so I removed it from all of them not knowing what they were for in the first place :)

Thanks for the explanation and I'll add those back in.

JPrevost:
@matt-bernhardt I think the latest commit address your analytics concern.

JPrevost:
Thanks for the review. I'm going to squash and merge.

app/helpers/bento_helper.rb
frrrances: duly noted.

PR #55, opened 2016-10-26
Add normalized fields

JPrevost:


JPrevost:
`html_safe` declares the content is safe... but actually doesn't make it safe. We can probably trust Google's blurbs to be safe, but I'm gonna fix that anyway.

Code reviews are good. Questions are good. Thanks!

:)

JPrevost:
@matt-bernhardt can you take a peek at that last change. Thx!

PR #57, opened 2016-10-27
Implements rudimentary summary bar

matt-bernhardt:
I'm probably doing unholy things with the ReportSummary function, but this is a starting point. I still need to make the items an internal link to the actual panels.
## Markup and Accessibility

One question for @frrrances : the markup I'm using for the summary bar is:

```
<div class="row bento-summary">
  <h3>Found</h3>
  <ul class="bento-summary-list">
      <li data-region="articles"><span class="count">491</span> Articles</li>
      <li data-region="books"><span class="count">5</span> Local Books</li>
      <li data-region="worldcat"><span class="count">7</span> Worldwide Books</li>
  </ul>
</div>
```

which renders for now as 
image

Some of this will be addressed in the UI work, but my thought here is to make screen reader output as simple as possible, like:

`List of three items. Item one 491 Articles. Item two 5 Local Books. Item three 7 Worldwide books` ... or some such - which is why I put the parentheses in via CSS rather than as characters.
## Application flow

@JPrevost the hope here is to populate each item in the summary bar as it comes back from the tool - rather than needing to wait until the end. I used the same `if box_enabled?()` construct in bento.html.erb that we use to toggle the various panels, and wrote a simple function called by _trigger_search.html.erb that handles the population. The `ReportSummary` function defines a default value, since I noticed that null results come back unpopulated.

The last piece to point out is that I added a data attribute to the markup via `search.html.erb`. This is what the javascript looks for to populate the summary bar.

### DIFF
--- a/app/assets/javascripts/static.js
+++ b/app/assets/javascripts/static.js
@@ -12,3 +12,8 @@ function TrackLinks( element ) {
     ga( 'send', fields );
   });
 }
+
+function ReportSummary( category, count ) {
+  count = ( typeof count !== 'undefined' ) ? count : 0;
+  $( '.bento-summary-list li[data-region="' + category + '"] .count' ).html( count );
+}
--- a/app/assets/stylesheets/bento.css
+++ b/app/assets/stylesheets/bento.css
@@ -9,3 +9,30 @@
 .bento-year {
   color: gray;
 }
+
+.bento-summary h3 {
+  display: inline-block;
+}
+
+.bento-summary ul {
+  display: inline-block;
+  padding-left: 0;
+}
+
+.bento-summary li {
+  display: inline-block;
+  padding-left: 1em;
+}
+
+.bento-summary li .count {
+  float: right;
+  padding-left: 0.25em;
+}
+
+.bento-summary li .count:before {
+  content: '(';
+}
+
+.bento-summary li .count:after {
+  content: ')';
+}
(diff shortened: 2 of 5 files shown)


frrrances:
I'm not super excited about putting the number in front of the type unless we remove the parentheses (and increase spacing which can come later). I think that would work best for both sighted and screenreader users.

frrrances:
Also, I know this is just a quick pass, but could we have the option of shortening for over 1000 to "k" (eg. 3000 would be 3k) or "1000+"?

JPrevost:
We can adjust the "friendliness" of the numbers at any time. This PR sets that in Rails which has some nice helpers for that.

@matt-bernhardt and I discussed the numbers before / after thing and he was thinking about screen readers when he put it first. I was cool with that as long as we can do some CSS to put it back after for non-screen readers :)

frrrances:
Otherwise, I like the list markup and general setup. 👍

matt-bernhardt:
Let me tweak the CSS rules to reverse the display order for visual user agents. I want to say `float: right` and a bit of margining will be enough to put things the way the mockup called for.

matt-bernhardt:
I took a listen to the results page in NVDA, and it seems like some more work is going to be needed on that front. The screen reader isn't always getting either the item counts, or the panel contents, in a logical fashion. There are ways to alert screen readers that content on a page has changed, but I'll need to read up on them before implementing anything.

For now, if @frrrances is comfortable with the markup/display as is, I can merge this - but I'll also log a separate issue in JIRA to look at screen reader output more generally.

JPrevost:
@matt-bernhardt the overall project plan includes getting IS&T to do an accessibility review. With that in mind, I think we should focus on matching our wireframes and adjusting markup later to address any usability concerns.

frrrances:
@matt-bernhardt - sounds good. As long as we stick with clean semantic markup, we can add eventing or adjust for screenreaders as we get closer.

PR #61, opened 2016-11-07
Basic styles to make results look real

frrrances:
This PR has some semi-quick styles to make the bento results look more like the wireframes. (The css will need to be cleaned up for prod but this will help with testing for now.)

Code review: 
@JPrevost or @matt-bernhardt 

Note on the codeclimate test failures: I'd like to talk about what's getting rejected before fixing.

*Things not included that still need to be done:*
- remove bootstrap and add in homegrown grid
- add summary text description for each box type
- adjust help blocks (content and layout)
- add jump links and smooth scroll for Found list
- add search term feedback summary below search box
- placeholder and spinner styling
- a11y review


JPrevost:
I can review this tomorrow. The coveralls test is helpfully pointing out that a method in a helper is no longer being used and can be deleted. Yay.

The css warnings in codeclimate, which are really css-lint in this case, seem questionable to me, but I'm not an expert in that domain by any means. Regardless, CodeClimate is not an automatic no-merge fail. If we don't agree with some of the rules, we can ignore / change them. If we do agree with them, but don't wanna fix it yet, we can come back to it later, etc.

app/views/search/_result.html.erb
```diff
@@ -1,43 +1,66 @@
...
-  <h3 class='bento-title'>
+<div class='wrap-result'>
+  
+  <% if result.type %>
+    <p class="result-type"><%= result.type %></p>
+  <% end %>
+
+  <% if result.thumbnail %>
+    <div class="result-image">
+      <%= image_tag(result.thumbnail) %>
```
frrrances: oh nice.

frrrances: 🌈

frrrances:
@matt-bernhardt  - yes i agree that the screen reader hasn't been optimized yet, nor have the semantics been set (a11y is in the still-yet-to-do column). This PR was meant to get the test env looking more like it will so we can start to evaluate data and results. I'll have a follow-up next week to address screen reader UX. (As for the list to paragraph, i was testing to see how smoothly it read. A list can come across as very burdensome when you have to enter the list and read off each item and i thought I would see if a single paragraph read more smoothly - not a decided point yet, just a test.)

PR #63, opened 2016-11-08
Updates encoding format for js triggered searches

JPrevost:
Why:

* Searches with apostrophes were being redirected as having invalid
  search targets. https://mitlibraries.atlassian.net/browse/DI-61

This change addresses the need by:

* Switching to `URI.encode_www_form_component` prevents javascript from
  choking on the passed parameters. Previously, with `URI.encode`
  javascript would lose all values in the query string after an
  apostrophe, including the target parameter which was what was
  causing the problems.

* We don't have any full stack integration tests for this type of work
  where we do a full search in the UI, therefore no test was added which
  is obviously not ideal. As the UI matures, we'll likely add some full
  stack javascript-enabled tests to help ensure we don't have
  regressions on this type of bug.


JPrevost:
@matt-bernhardt can you do a quick review on this?

PR #66, opened 2016-11-10
Adds jumpstart search links for author and subject

JPrevost:
What:

* This commit provides jumpstart links into appropriate EDS
  profiles to allow users to start searches for an author or
  subject they click on in the bento interface.

Why:

* Users may want to run searches on specific authors or subjects
  they find in our bento search in order to find more results
  that are similar or related.
* https://mitlibraries.atlassian.net/browse/DI-95
* https://mitlibraries.atlassian.net/browse/DI-111

How:

* The normalized authors will now contain an array of arrays. The inner
  arrays will be ['Author Name', 'Link to jumpstart search'].
* The work of creating appropriate links is done in the normalizer so
  we can keep the views as clean as possible and make the logic more
  easily testable with unit tests.
* Each EDS profile has an ENV that sets the base of the URL, we then
  follow the example we see in the EDS UI itself in which to search
  for an author from a record they do a keyword search for:
  `AU "author name"` so we append that to the base URL.
* For Subject searches we append: `DE "subject string"`.

### DIFF
--- a/app/models/normalize_eds.rb
+++ b/app/models/normalize_eds.rb
@@ -70,7 +70,7 @@ class NormalizeEds
   def construct_open_url(record)
     'https://sfx.mit.edu/sfx_local?' +
       NormalizeEdsArticles.new(record).openurl(
-        record, year(record), authors(record)
+        record, year(record), authors(record).map(&:first)
       )
   end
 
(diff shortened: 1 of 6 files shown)


JPrevost:
@matt-bernhardt please review if you have time. Thx!

JPrevost:
Thx @matt-bernhardt (and yes, I agree. We'll want to figure out how many subjects, etc we want to display... probably do something like we do with authors).

PR #67, opened 2016-11-10
Adds link to sfx for journals

JPrevost:
What:

* This commit provides links into SFX for journals in our
  articles search bento box

Why:

* Users may want to find more information about a journal rather
  than linking to the specific article we have returned in the
  search results
* https://mitlibraries.atlassian.net/browse/DI-112

How:

* The normalized `in` will now contain an array of Journal Title
  and sfx link, like `['Journal Title', 'link to SFX for Journal']`
* The work of creating appropriate links is done in the normalizer so
  we can keep the views as clean as possible and make the logic more
  easily testable with unit tests

### DIFF
--- a/app/models/normalize_eds_articles.rb
+++ b/app/models/normalize_eds_articles.rb
@@ -10,7 +10,7 @@ class NormalizeEdsArticles
     {
       'rft.au' => authors&.join(';'),
       'rft_id' => "info:doi/#{doi(record)}",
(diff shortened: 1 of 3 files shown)


JPrevost:
@frrrances or @matt-bernhardt if either of you have time for a review that would be great! Thx.

JPrevost:
Thx @matt-bernhardt

PR #69, opened 2016-11-15
Prep for bootstrap removal

JPrevost:
What:

* Removes the reliance on the bootstrap gem's flash helper
* Moves some logic of the application layout template into partials
* Updates dependencies

Why:

* To help make moving to a custom grid easier, this work removes our
  reliance on the bootstrap gem's flash helper method. Currently, this
  only supports `error` flashes, but that's also all this app actually
  uses
* The html is still what bootstrap uses. That may change later
* The application layout template was awful to work with so I moved
  sections into partials

See Also:

* The bootstrap feature we will want to replicate is detailed here:
  http://getbootstrap.com/components/#alerts
* The bootstrap_flash_helper is here in case we want to use it as a
  basis for supporting all of the flash types eventually:
  https://github.com/seyhunak/twitter-bootstrap-rails/blob/master/app/helpers/bootstrap_flash_helper.rb


JPrevost:
@frrrances Can you review this? If moving to partials for the main layout sections makes your current work a mess, I can do that work later. I think it'll help with sanity longterm though.

The main goal with this is to allow us to remove all bootstrap remnants soon and the use of the flash helper was likely to make that more complicated than just removing the gem and changing our css.

Also, you can see the alert by running a search with no search term added.

JPrevost:
Also, because a lot of this was just copying stuff to the new partials the main bit to review is probably just the new flash partial (I should have done two PRs... one for new stuff, one for the moves. Sorry).

frrrances:
yay for moving things into partials! and for disentangling bootstrap. 👍

PR #74, opened 2016-11-21
Handle different eds title mappings

JPrevost:
What:

* Records in EDS don't have a consistent way of signifying a title that
  is useful for all records. This attempts to handle two common variants
  while falling back to a default "unknown title" if neither match.
* This also include a refactoring to split EDS common normalization into
  a new class to prevent the main class from getting overly complicated.

Why:

* We need to handle different ways of mapping information from EDS
  records to allow us to present the best information possible to our
  users.
* https://mitlibraries.atlassian.net/browse/DI-122


JPrevost:
@matt-bernhardt or @frrrances : can one of you review this? Maybe comment back here if you can do a review? Thx!

JPrevost:
I'm gonna merge. If y'all end up with concerns on this code let me know and I'll address them.

PR #75, opened 2016-11-22
Limit long subject lists

JPrevost:
What:

* creates a convenience method for limiting display to 3 subjects

Why:

* some records have _lots_ of subjects and it is unlikely it will be
  useful to display them all
* https://mitlibraries.atlassian.net/browse/DI-127


JPrevost:
@frrrances please review and confirm this is what we are intending to do. Thanks!

frrrances:
perfecto! 👍

PR #76, opened 2016-11-22
Adds support for direct record url

JPrevost:
What:

* This allows for articles to have distinct links for detailed record
  view and get_it_url

Why:

* it was decided to always have the item title link to the record view
  in the source while having the get_it link go to either the record
  view or sfx depending on the type of item
* https://mitlibraries.atlassian.net/browse/DI-128

How:

* The `Result` class now supports both the required `url` as well as a
  new optional `get_it_url`. The normalizers set those as appropriate
  for the different types of resources.

### DIFF
--- a/app/models/normalize_eds_articles.rb
+++ b/app/models/normalize_eds_articles.rb
@@ -8,7 +8,28 @@ class NormalizeEdsArticles
   def article_metadata(result)
     result.citation = numbering
     result.in = journal_title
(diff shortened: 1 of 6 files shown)


JPrevost:
@frrrances Let me know if this does what y'all were expecting. Thx!

frrrances:
Yup, looks good to me! 👍

PR #77, opened 2016-11-28
Adds db_source display

JPrevost:
What:

* adds db source label and db source id to results from EDS sources

Why:

* While we assess what data sources should feed each box in the bento,
  it will be helpful to know what data sources the specific results we
  are seeing come from
* https://mitlibraries.atlassian.net/browse/DI-130

### DIFF
--- a/app/views/search/_result.html.erb
+++ b/app/views/search/_result.html.erb
@@ -61,6 +61,12 @@
       </div>
     <% end %>
 
+    <% if result.db_source.present? %>
+      <p class="result-pubinfo">
+        Source DB: <%= result.db_source[0] %> (<%= result.db_source[1] %>)
+      </p>
+    <% end %>
+
     <% if result.url %>
       <div class="result-get">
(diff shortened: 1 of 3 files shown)


app/views/search/_result.html.erb
```diff
@@ -61,6 +61,12 @@
       </div>
     <% end %>
 
+    <% if result.db_source.present? %>
+      <p class="result-pubinfo">
```
JPrevost: I used an existing class because this is kinda temp and I know I'd forget to remove any custom CSS classes I made when this gets axed.

JPrevost:
@frrrances can you pop in for a quick review of this?

frrrances:
Looks good. Do you want to add a TODO: anywhere in there? Up to you on that... 👍  either way.

JPrevost:
Nah. We'll get sick of it eventually and someone will tell us to remove it entirely ;) Thx!

PR #80, opened 2016-12-09
Adds global alert system

JPrevost:
What:

* Adds an ENV based global alert system

Why:

* Initially, this will be used to signify test and beta instances of our
application. Eventually, this may be used to signify other globally
relevant application level information (such as known planned API
outages, etc)
* See: https://mitlibraries.atlassian.net/browse/DI-141

How:

* ENV['GLOBAL_ALERT'] can be set like 'This is a beta site. You may be
interested in our
<a href="/some_url">production discovery environment.</a>'
* By using ENV['GLOBAL_ALERT'] we can always set our test ENVs to link
back to Prod versions, etc

### DIFF
--- a/app.json
+++ b/app.json
@@ -41,6 +41,9 @@
     "ENABLED_BOXES": {
       "required": true
     },
+    "GLOBAL_ALERT": {
+      "required": true
+    },
     "GOOGLE_API_KEY": {
       "required": true
     },
--- a/app/views/layouts/_flash.html.erb
+++ b/app/views/layouts/_flash.html.erb
@@ -1,5 +1,8 @@
-<% if flash[:error] %>
-  <div class="alert alert-banner alert-danger" role="alert">
-    <%= flash[:error] %>
-  </div>
+<% if flash %>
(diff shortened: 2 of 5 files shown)


frrrances:
@JPrevost - Here is a PR with a static banner: https://github.com/MITLibraries/bento/pull/81

JPrevost:
1. At this time, someone from DLAD will indeed have to manage global alerts. If that proves to be problematic, an admin interface with a db backed alert could be implemented. Only one global alert is currently supported and I feel that is reasonable for the intent of global alerts. It is of course possible that someone will come up with a reason to need more, but those could probably be handled with a single alert as it can handle embedded html.

2. The flash alerts are more for normal application flow level alerts. Right now, there is only one and that is for when a search term is not entered. Those can indeed handle displaying multiple alerts at one time, even if the application doesn't currently use that. There are caveats around this in that while you can display multiple alerts via the flash system, only one per type will survive. It generally ends up being fine.

matt-bernhardt:
Ok, cool - thanks for answering those questions.

PR #87, opened 2016-12-16
Di144 bye bye whatnot

JPrevost:
What:

* Removes the temporary whatnot box to move us closer to our test
target. https://mitlibraries.atlassian.net/browse/DI-144
* Adds box descriptions and creates place to put the help stuff and
the other resources links. All of those needs styles and some of it
needs markup. https://mitlibraries.atlassian.net/browse/DI-145
* removes some remnants of using multiple EDS profiles that were
confusing me


JPrevost:
The other box uses dl/dt/dd here. That can change.

Sorry for throwing in the session box removal thing on this pr but it was driving me nuts.

JPrevost:
list stuff is good too. Merging.

PR #89, opened 2016-12-16
make global notice dark

frrrances:
it was too light.


frrrances:
Since this is an "informational" notice (class="info"), the style is laid back (e.g. grey). We can switch it to "warning" type (class="warning") and it will be yellowish. I would prefer to save red for "error" types that indicate major problems or irreversible changes.

JPrevost:
Yeah, I think the question of "you are probably on the wrong server" of being information or not is one we can debate later on.

frrrances:
... or is it the right server? food for debate.

PR #90, opened 2016-12-19
Updates Dependencies

JPrevost:
What:

* updates gem dependencies to latest versions
* updates a few methods with empty bodies to match the newly enabled
  default rubocop rule

Note:

* all project developers will need to run `bundle install` to update
  local branches to these updated dependencies


JPrevost:
@frrrances and @matt-bernhardt this doesn't really need a review, but I'm tagging y'all so you can see the note about how you'll need to run `bundle install` again to grab these new dependencies once this is merged. Also, rebase any open branches once this is merged right away to avoid having different sets of dependencies on different branches... technically it works just fine, but it's best to just work off the new deps as soon as possible.

PR #91, opened 2016-12-19
add about bento block

frrrances:
This PR adds an "About bento" informational block to the bottom of the bento pages. 
NOTE: Copy is still TBD but shippable.


JPrevost:
:shipit:

PR #93, opened 2016-12-20
View more pagination and style adjustments

frrrances:
This PR adds pagination styles and breadcrumbs, and adjusts some visual styles (more boxiness and colors).


frrrances:
Yes, I should probably start doing some documentation for that stuff (which lives on the style guide which I also need to move over to MitLib githu). I'll ping you when I have something. 

(If you want the short answer, the css classes for any layout band are structured as 
```layout-[columns]```
where the columns are for large screens. Most just bump down to single column when it makes sense to. The two layouts there are 1/3-2/3 (1-third 2-thirds) and 1/4-1/2-1/4 (1-quarter 2-quarters 1-quarter). You can find the others in the layout partial file.

PR #95, opened 2017-01-09
Adds sidebar to view more results views

JPrevost:
* adds page heading and sidebar to view more results views
* see also https://mitlibraries.atlassian.net/browse/DI-151

Note: this is a bit kludgy and needs to styles work, but should allow us
to confirm this is what we want


JPrevost:
@frrrances All links are currently active in the sidebar but wanted to confirm this is the gist of what we're looking for

PR #101, opened 2017-01-23
Di178 view more toggle

JPrevost:
What:
* Allows a toggle between viewing local paginated results and results
  in EDS UI
* The default is EDS UI
* This is a global setting
* To enable local paginated results, set Environment Variable
  `LOCAL_RESULTS=true`

Why:
* It is still not clear whether we want to send people to EDS UI or use a
  local solution for this feature
* This feature toggle will allow us to test both options to make an
  informed decision on what is best
* See https://mitlibraries.atlassian.net/browse/DI-178


JPrevost:
fwiw, that code climate brakeman flag is a false positive.

PR #103, opened 2017-01-24
Di59 result error handling

JPrevost:
What:

* This presents an error message to users when things go awry during the
  XHR requests
* This does not try to get fancy and report errors from the client side,
  but instead assumes the errors would have been logged during the rails
  server side processing (which is probably true)
* It does log to the end user's JavaScript console the xhr response when
  this is triggered so in theory we could work with tech savvy users to
  get a copy of the request if necessary (probably not)
* This also adds a timeout of 6 seconds to the EDS API calls and logs
  all URLs to EDS for debug purposes
* Disables rollbar and enables more logging in dev environment

Why:

* The previous fail state was an infinitely spinning loader image which
  was obviously not too helpful

Note:
* see also https://mitlibraries.atlassian.net/browse/DI-59

### DIFF
--- a/app/models/search_eds.rb
+++ b/app/models/search_eds.rb
@@ -46,9 +46,13 @@ class SearchEds
   end
 
   def search_filtered(term, facets, page, per_page)
+    Rails.logger.debug(
+      "EDS Search URL: #{search_url(term, facets, page, per_page)}"
+    )
     result = HTTP.headers(accept: 'application/json',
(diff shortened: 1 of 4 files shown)


JPrevost:
@matt-bernhardt I flagged you for review as this affects the GA logging you did. I think I didn't break it, but I also didn't check ;)

frrrances:
UX lgtm 👍

PR #105, opened 2017-01-25
Text changes and hide source (for now)

frrrances:
This PR makes a few small changes in response to the feedback from testing (and in prep for the next round of testing).

### DIFF
--- a/app/assets/stylesheets/bento.scss
+++ b/app/assets/stylesheets/bento.scss
@@ -397,3 +397,6 @@ $z-depth-way-front:     1000;
   font-weight: $fw-bold;
 }
 
+.wrap-result>.result-pubinfo {
+  display: none;
+}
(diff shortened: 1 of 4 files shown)


app/views/search/bento.html.erb
```diff
@@ -20,7 +20,7 @@
 </div>
 
 <div class="gridband layout-2c">
-  <%= render partial: "placeholders", locals: { heading: 'Website and guides', id: 'website_content', description: 'MIT Libraries website and research guides.' } %>
+  <%= render partial: "placeholders", locals: { heading: 'Library website and guides', id: 'website_content', description: 'MIT Libraries website and research guides.' } %>
```
JPrevost: Do you want to update the `results-summary` wording to match this change?

app/assets/stylesheets/bento.scss
```diff
@@ -397,3 +397,6 @@ $z-depth-way-front:     1000;
   font-weight: $fw-bold;
 }
 
+.wrap-result>.result-pubinfo {
```
JPrevost: We should probably create a debug mode that re-enables this and maybe some other stuff. But this is fine for now.

PR #117, opened 2017-02-03
Recaptcha feedback form

JPrevost:


JPrevost:
The warning on the PR build heroku site is expected since we are using test keys there (it always passes). Once moved to stage/prod it'll work with our site keys as expected.

PR #119, opened 2017-02-06
Add Recaptcha to Feedback Form

JPrevost:
What
- adds recaptcha to feedback form
- uses test keys in pr build

Why
- to prevent spam

### DIFF
--- a/app.json
+++ b/app.json
@@ -56,6 +56,8 @@
     "RAILS_SERVE_STATIC_FILES": {
       "required": true
     },
+    "RECAPTCHA_SECRET_KEY": "6LeIxAcTAAAAAGG-vFI1TnRWxMZNFuojJ4WifJWe",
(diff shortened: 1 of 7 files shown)


app.json
```diff
@@ -56,6 +56,8 @@
     "RAILS_SERVE_STATIC_FILES": {
       "required": true
     },
+    "RECAPTCHA_SECRET_KEY": "6LeIxAcTAAAAAGG-vFI1TnRWxMZNFuojJ4WifJWe",
```
frrrances: is this ok to have in a public repo?

JPrevost: Thanks for checking! But yes, those are special "test keys" google provides for this purpose.

frrrances: ah that explains why it let me off so easy. clearly i am a robot.

JPrevost: :)

More info if you care:
https://developers.google.com/recaptcha/docs/faq

PR #123, opened 2017-02-06
Better analytics

matt-bernhardt:
This changes our analytics approach to achieve two goals:
1) Recording of additional interactions (increased scope)
2) Recording of relevant context (less detail)

### DIFF
--- a/app/views/layouts/_head.html.erb
+++ b/app/views/layouts/_head.html.erb
@@ -41,7 +41,7 @@
 
   // Interaction tracking
   $(document).ready(function(e) {
-    TrackLinks( $( '.region a' ) );
+    TrackLinks( $( 'a' ) );
   });
 
 </script>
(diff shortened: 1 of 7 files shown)


matt-bernhardt:
image

This is how the events are being recorded in Google Analytics

app/views/layouts/_head.html.erb
```diff
@@ -41,7 +41,7 @@
 
   // Interaction tracking
   $(document).ready(function(e) {
-    TrackLinks( $( '.region a' ) );
+    TrackLinks( $( 'a' ) );
```
JPrevost: This will track all links with Events, internal and external... will that result in double counting stats? (one for the normal GA code and one for the Event?). I think it's probably okay, I just want to make sure we aren't going to mislead ourselves in terms of usage later on.

matt-bernhardt: The expansion will end up tracking all interactions with links, yes - so in the case where a link also results in a page load within Bento, that page load will be counted as part of the page load approach as well as recorded as an Event. IMO, this is only problematic if we take the step of adding event counts and page load counts, which I would argue is bad practice even without this step.

PR #129, opened 2017-02-10
Use same number format in heading we do in view more

JPrevost:


JPrevost:
@matt-bernhardt ... this may affect the analytics bits you are working on so I wanted to call it to your attention explicitly

PR #134, opened 2017-02-22
TEMP: hide status, and add temp links to view more sidebar

frrrances:


app/views/search/_result.html.erb
```diff
@@ -63,7 +63,7 @@
       <ul class="list-local-locations">
       <% result.location.each do |location| %>
         <li class="result-local-location">
-          <%= location[0] %>: <%= location[1] %> (not live)
```
JPrevost: This line shows us the cached status from EDS. It is displayed if anything goes wrong when requesting the realtime information. The text should reflect that it is not realtime information. I don't particularly care what exactly we say, but we should make it clear it is not realtime.

app/views/search/search_boxed.html.erb
```diff
@@ -40,12 +40,12 @@
...
     <div class="bit">
-      <h3>Supporting content</h3>
-      <p>Lorem ipsum dolor sit amet, consetetur sadipscing elitr: </p>
+      <h3>Other options</h3>
+      <p>If you're not finding what you need, you can try one of these: </p>
       <ul>
-        <li class=""><a href="">Sed diam nonumy eirmod</a></li> 
-        <li class=""><a href="">Tempor invidunt ut labore</a></li> 
-        <li class=""><a href="">Dolore magna aliquyam</a></li>
+        <li class="http://libraries.mit.edu/bartonplus"><a href="library.mit.edu">Search in BartonPlus</a></li>
```
JPrevost: This css class is probably not what you intended ;)

Also, is there a reason we don't consistently use https when linking to libraries.mit.edu?

frrrances: ... and this is why we have code reviews *before* merge. 😞  and yes, we should be linking with https. I'll make any changes in the next PR.

PR #136, opened 2017-02-28
Adds Realtime Status Icons

JPrevost:
What
- adds realtime status icons

Why
- quick visual indicator of whether something is immediately available
- see also: https://mitlibraries.atlassian.net/browse/DI-239


app/assets/stylesheets/aleph.scss
```diff
@@ -0,0 +1,11 @@
+// Place all the styles related to the aleph controller here.
+// They will automatically be included in application.css.
+// You can use Sass (SCSS) here: http://sass-lang.com/
+
+.fa-check {
+  color: green;
```
frrrances: I'd like to use the color variables here: $success and $error (i know it's not an error but hey). This means we'll have to get this in the right area to know the variables exist. Let me know if you need help with that (or you can put this in bento.scss)

app/models/aleph_item.rb
```diff
@@ -22,8 +22,32 @@ def items(id)
...
+      call_number: item.xpath('z30/z30-call-no').text,
+      available?: available?(item),
+      label: label(item) }
+  end
+
+  def status(item)
+    item.xpath('status').text
+  end
+
+  def available_statii
```
frrrances: ooo statii.

app/assets/stylesheets/aleph.scss
JPrevost: Additional commit addresses this. We'll now have to (get to?) explicitly include each css/scss file in the order that makes sense.

PR #137, opened 2017-03-02
Add "displaying k1-k2 of n" indicator to results view

thatandromeda:


app/views/search/search_boxed.html.erb
```diff
@@ -7,11 +7,14 @@
         articles and journals
       <% end %>
 
-      for 
+      for
 
       <span class="search-string">"<%= params[:q] %>"</span>
     </h2>
     <p class="subline">
+        <%= page_entries_info @pageable_results, entry_name: 'result' %>, sorted by relevance
```
JPrevost: This is super minor as it falls into the "I don't care much about the html templates as they are always a mess anyway" rule... but Rails convention is two space indents not this god awful python four space indent ;)

thatandromeda: Don't suppose you know a way to make tab size in Atom sensitive to filetype?

JPrevost: Uh, I don't... but mine does. Weird.

app/controllers/search_controller.rb
```diff
@@ -17,10 +17,7 @@ def search
     @results = search_results(page, per_page)
     # EDS API Limits retrieving results past about 250
     max = [@results['total'], 250].min
-    @pageable_results = Kaminari.paginate_array(
-      @results['results'],
-      total_count: max
-    ).page(params[:page]).per(10)
+    @pageable_results = paginate_results(@results, max, params)
```
JPrevost: I hope you find my comments helpful as you learn the Rails Way. I'm being nit picky to point out things that are different than python and/or 

@results is already in scope to all methods in the controller so you don't need to pass it. As is params... and maybe move the max calculation to the private method as it's only used now down there.

I think we'll want to pass in `page` and `per_page`. I think not doing might have been the cause of this bug:
https://mitlibraries.atlassian.net/browse/DI-242

Or, more accurately, why we were _seeing_ that bug. The ticket should remain open so we handle it gracefully, but I think passing per_page as calculated in the method will prevent us from creating links that go too far. (this clearly screams for a better test, but we can handle that with that Jira ticket).

Finally,  Ruby automatically passes the last value so you almost never have to explicitly `return` anything.

Here's where I'd land with that method:
```
  # Turns search results into a format that can be used by the paginator.
  def paginate_results(page, per_page)
    max = [@results['total'], 250].min
    Kaminari.paginate_array(
      @results['results'],
      total_count: max
    ).page(page).per(per_page)
  end
```

Please feel free to ask questions if my comments are not helpful either here or on Slack.

JPrevost:
This is all good to squash and merge. :shipit: 

🎉

PR #138, opened 2017-03-02
Expands options and prioritizes of links

JPrevost:
What:

* Extracts an additional link of interest
* Prioritizes "best" link as `getit_url`. This is likely to need
  adjustment over time

Why:

* The additional link we are extracting is the best link available
  for many records


JPrevost:
@frrrances I merged before review because functionally this has been signed off on as being better than what we had before. When you get back, it would be great to have a review anyway :) Thanks!

frrrances:
lgtm! 👍

PR #148, opened 2017-03-09
Wrap 404s in site template

JPrevost:
This just puts the 404s in our site template for now.

5xx is less simple due to the nature of 5xx meaning it's best to use static pages so we need to do more work.


app/views/catch_all/catch_all.html.erb
```diff
@@ -0,0 +1,3 @@
+<h2>Ooops! The requested page was not found.</h2>
```
frrrances: Can you add ``class="title title-page" `` to this h2?

JPrevost: 👍

PR #156, opened 2017-03-17
Removes Record Links from GetIt Link

JPrevost:
What:

* Removes record links (eds ui detail view) from GetIt link
* Removes constructed openurls from GetIt link
* Extracts both fulltext and record links from EDS
* Adds record link as a button (only using EDS UI link for now)
* Renames GetIt Link to reflect it should lead towards online
  full text links
* Splits record debug to a partial to make it easier to scan the main
  record view

Why:

* The single button doing so much was potentially leading to confusion
for end users

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -6,3 +6,4 @@
 @import "bento.scss";
 @import "aleph.scss";
 @import "feedback.scss";
+@import "results.scss";
--- a/app/views/search/_result.html.erb
+++ b/app/views/search/_result.html.erb
@@ -91,48 +91,14 @@
   <% end %>
 
   <div class="result-get">
-    <%= link_to(result.get_it_label, result.getit_url, class: 'button button-primary', data: {type: "Get"}) %>
+    <% if result.getit_url.present? %>
+      <%= link_to("Online", result.getit_url, class: 'button button-primary green', data: {type: "Get"}) %>
+    <% end %>
+
+    <%= link_to("Details and requests", result.url, class: 'button button-secondary', data: {type: "Detail"}) %>
   </div>
 
   <% if session[:debug] %>
-    <p class="result-debug">
-
-      <ul>
-        <li>EDS Supplied CustomLinks</li>
-          <ul>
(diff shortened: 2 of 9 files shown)


JPrevost:
Note: this needs some better wording for the buttons and better styles, but should serve as a place to discuss this feature and move it forward

app/assets/stylesheets/application.scss
```diff
@@ -6,3 +6,4 @@
 @import "bento.scss";
 @import "aleph.scss";
 @import "feedback.scss";
+@import "results.scss";
```
frrrances: is there a reason you broke this out into it's own partial? I'd prefer the code in it go with the other result styling, so I can move it all here (if the compile is allowing variables) or I can move the new code into bento. Let me know what you prefer, or if you want to fight about it.

app/views/search/_result.html.erb
```diff
@@ -91,48 +91,14 @@
...
-          <%= link_to("MARC 856", result.marc_856, class: 'bento-link') %>
-        </li>
-        <% else %>
-          <li>No MARC 856</li>
-        <% end %>
-      </ul>
-
-      <ul><li>Winner: <%= result.winner %></li></ul>
-    <p>
+    <%= render partial: "debug", locals: { result: result } %>
```
frrrances: nice cleanup.

app/assets/stylesheets/application.scss
JPrevost: Since we're really only using this style on results-y things it's easier to know where you're using it if you put it in a sheet named the same as the controller/views where it's used. My preference would be for the styles in bento to be split into different sheets unless they are actually used in multiple places... but that may be worth more discussion. Mostly I find it impossible to ever delete styles if they all get thrown into a shared sheet because you never know if they are used someplace else.

frrrances: I totally agree and usually do it that way but our original compile wasn't letting us use variables from other sheets - let me see if we can use them here.

frrrances: 👍

PR #163, opened 2017-03-24
Numbers are not strings

JPrevost:
- this fixes a bug where the logic to determine whether to show a view more link
  broke if `RESULTS_PER_BOX` was set


JPrevost:
The heroku app associated with this build as `ENV['RESULTS_PER_BOX']=3` set to allow confirmation this is working as intended with the value set. Normally PR builds skip that setting because it is optional. We set it in staging and production which is how this bug was noticed.

PR #166, opened 2017-03-30
View all links feature toggle

JPrevost:
What:

* moves all feature toggles to shared route. debug used to be special,
  but now that we will have more than one toggle, a generic feature
  toggle route that accepts feature parameters is used instead.
  ex. instead of `/debug` the new process is `/toggle?feature=debug`
* configures view all links to use either local or remote browsing
* due to the nature of the query caching, it was necessary to always
  provide both links in the cacheable response so users with one
  preference don't end up giving default users the cached links for the
  other preference. the only real downside to this is it will be
  slightly more work to clean up when we get rid of the remote option

Note:

* There are outstanding bugs and missing features relating to local
  browsing noted as related tickets here:
  https://mitlibraries.atlassian.net/browse/DI-260


app/views/search/_view_all_link.html.erb
```diff
@@ -2,9 +2,14 @@
   <div class="results-more">
   <% if @results['total'] > 0 %>
     <% return unless @results['total'] > @per_box %>
+    <% if Flipflop.enabled?(:local_browse) && @results['local_view_more'] %>
```
frrrances: Does it make sense to pull this logic out into a feature setting area? As we add more pieces that get toggled (I'm thinking of local browse for view all, subject, and author links at this point), we'll have logic scattered across several files.

app/controllers/feature_controller.rb
```diff
@@ -0,0 +1,45 @@
+class FeatureController < ApplicationController
+  before_action :validate_feature!
+
+  # Toggles boolean for session based feature flags
+  def toggle
```
frrrances: Would it be possible to have a way to check the status of any of the feature toggles as a batch and display a message? Something like ``/toggle?feature=report`` - I'm guessing if we have this enabled on staging we'll set something and forget and have to debug the changed behavior at some point.

JPrevost: Sure. Ticket!

The built in dashboard may be worth implementing instead of rebuilding it ourselves though. I'm thinking of something simple like making the dashboard protected by a query_param with a shared secret we link from our non-public project documentation to allow project staff to enable / disable / check status on their session for all feature toggles. Thoughts?

frrrances: If that's easier, I'm all for it. Sounds kind of like a Cadillac! You want me to make a ticket for that option? Or just a general one and you can decide what is most efficient?

app/views/search/_view_all_link.html.erb
JPrevost: Yeah. Good idea. 🚮

I had to take the logic out of the model because of the cache, but now that you called me out on it I have some ideas of how to clean up the views while still not causing cacheing issues. I think I'll handle that refactor when I work on the related features / bug fixes.

app/controllers/feature_controller.rb
JPrevost: A general ticket is cool so I don't forget but don't want to hold off on the merge if that's cool.

app/views/search/_view_all_link.html.erb
frrrances: fine.

app/controllers/feature_controller.rb
frrrances: 👍

PR #171, opened 2017-04-04
288 nomethod error

JPrevost:
Metadata with no relationships when reading year
* metadata that came back from EDS that did not contain
  `IsPartOfRelationships` would fail when trying to extract the year
  from that section
* this handles that case by confirming the expected sections exist
  before trying to access them

See https://mitlibraries.atlassian.net/browse/DI-288

This also adds an option to log like production during local development.

### DIFF
--- a/app/models/normalize_eds_common.rb
+++ b/app/models/normalize_eds_common.rb
@@ -60,7 +60,7 @@ class NormalizeEdsCommon
   end
 
   def year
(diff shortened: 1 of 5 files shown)


JPrevost:
Heroku is being weird. The app built but it isn't showing up in this PR...
https://mit-bento-staging-pr-171.herokuapp.com/

PR #177, opened 2017-04-10
Allows ENV configurable http_timeout in EDS Search

JPrevost:
What:

* it is now possible to control the timeout via an Environment
  setting for EDS API calls

Why:

* there are times with EDS doesn't seem to be responding in an
  acceptable amount of time
* while we feel EDS should fix this, it is likely there will be
  times when we want to bump our timeout a bit higher just to allow
  users to search even if it takes way too long

### DIFF
--- a/app/models/search_eds.rb
+++ b/app/models/search_eds.rb
@@ -49,9 +49,26 @@ class SearchEds
     result = HTTP.headers(accept: 'application/json',
                           'x-authenticationToken': @auth_token,
                           'x-sessionToken': @session_key)
-                 .timeout(:global, write: 2, connect: 2, read: 2)
+                 .timeout(:global, write: http_timeout,
+                                   connect: http_timeout,
+                                   read: http_timeout)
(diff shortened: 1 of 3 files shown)


app/models/search_eds.rb
```diff
@@ -49,11 +49,28 @@ def search_filtered(term, facets, page, per_page)
...
+  # is possible to instead separate out each phase of the request for
+  # distinct timeouts if we find that is better.
+  # https://github.com/httprb/http/wiki/Timeouts
+  def http_timeout
+    t = if ENV['EDS_TIMEOUT'].present?
+          ENV['EDS_TIMEOUT'].to_f
+        else
+          6
+        end
+    (t / 3)
```
frrrances: are you asking for a number and then dividing it by 3?

JPrevost: lol. Yup. Does my note make this more or less confusing about what is happening?

frrrances: ha! naturally i didn't read it. it *all* makes so much sense now.

PR #180, opened 2017-04-13
Realtime avail and Link Tracking for Local browse

JPrevost:
What:

* adds realtime availability and ga link tracking when using the local
  view more feature

Why:

* it was a bug it wasn't working

How:

* added parent div to results like we have in bento view so we can
  target results with javascript functions

See: https://mitlibraries.atlassian.net/browse/DI-243

### DIFF
--- a/app/views/search/bento.html.erb
+++ b/app/views/search/bento.html.erb
@@ -36,7 +36,5 @@
 
   <%= render partial: "trigger_search", locals: { target: 'books', id: 'books_content'} %>
 
-  <%= render partial: "trigger_search", locals: { target: 'whatnot', id: 'whatnot_content'} %>
-
   <%= render partial: "trigger_search", locals: { target: 'google', id: 'website_content'} %>
 </script>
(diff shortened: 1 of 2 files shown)


app/views/search/bento.html.erb
```diff
@@ -36,7 +36,5 @@
 
   <%= render partial: "trigger_search", locals: { target: 'books', id: 'books_content'} %>
 
-  <%= render partial: "trigger_search", locals: { target: 'whatnot', id: 'whatnot_content'} %>
```
matt-bernhardt: I must have missed something along the way here - I don't see a whatnot panel, either in production or in this branch - so I'm not sure what this changes?

This isn't a blocking comment, more a "what did I miss" question.

frrrances: We dropped the whatnot profile (recently) and box (a while ago) so i think this was just clean up.

JPrevost: yup. cleanup. 🛀

PR #182, opened 2017-04-24
Adds in-app subject and author links

JPrevost:
* local_browse feature now uses in_app links instead of external links
  to EDS UI for authors and subjects
* external links are still used when local_browse is not enabled

NOTE: I'm not sure this is actually working as intended as I don't seem to get identical results browsing locally and via the EDS UI options.

HINT: https://mit-bento-staging-pr-182.herokuapp.com/toggle?feature=local_browse

### DIFF
--- a/app/models/normalize_eds_books.rb
+++ b/app/models/normalize_eds_books.rb
@@ -16,12 +16,16 @@ class NormalizeEdsBooks
 
   def subjects
     bibrecord['BibEntity']['Subjects']&.map do |s|
-      [s['SubjectFull'], subject_link(s['SubjectFull'])]
+      [subject_name(s), subject_link(s)]
     end
   end
 
+  def subject_name(subject)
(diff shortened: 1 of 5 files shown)


frrrances:
I'm noticing a few things: 
* Author search seems to work correctly, but not use the filters to limit format
* Subjects don't seem to be working entirely correctly, though I can't figure out a pattern yet - [example](https://mit-bento-staging-pr-182.herokuapp.com/search/bento?q=AU+%22Adams%2C+Douglas%22) that shows results when you click "electronic books" but not when you click "Prefect, Ford -- Fiction" even though 2 of the top 5 results have that subject heading.
I'll keep digging but I think you're right that this needs some tweaking.

JPrevost:
I'm in no rush to merge. We can send this out for more feedback too.

frrrances:
Subjects looking good now  🌈

PR #184, opened 2017-04-25
Change online button to action

frrrances:
This PR changes the wording on the online button to make it more obvious that it is a button that gets the online thing.


frrrances:
relates to #185

PR #193, opened 2017-05-08
Keep view more link in dom when not displaying

JPrevost:
* we were not adding the view more link when we had fewer results in
  the box than the per_box max which prevented the data elements from
  being present for the javascript to use to populate the summary bar
* this adds the view more link back to the dom and adds an `hidden`
  css class
* one day, Frances promises she will make that class something useful

https://mitlibraries.atlassian.net/browse/DI-302


JPrevost:
I approve this PR 👍

PR #194, opened 2017-05-09
Flag title values as html_safe because they are

JPrevost:
What:

* flag title values as html_safe so they won't be escaped and thus
  display properly.

Why:

* nobody wants to see `&amp;`

See: https://mitlibraries.atlassian.net/browse/DI-329


JPrevost:
@frrrances well, they are safe as in "not user supplied content"... do you trust Ebsco? That's the real question here :)

PR #217, opened 2017-06-27
Hint styles

frrrances:
This PR adds some hint styles in case we need them for surfacing DBs in the short term.

NOTE: The HTML example commit should be removed before merge.

### DIFF
--- a/app/assets/stylesheets/_hint.scss
+++ b/app/assets/stylesheets/_hint.scss
@@ -1,0 +1,26 @@
+
+.hint-box {
+  position: relative;
+  margin: 1.5% 0;
+  padding: 1.5%;
+  background-color: $wisp;
+
+  .type {
+    position: absolute;
+    top: 1rem;
+    right: 1rem;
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/_hint.scss
```diff
@@ -0,0 +1,28 @@
...
+		font-size: $fs-small;
+		font-weight: $fw-bold;
+	}
+
+	.desc {
+		margin-bottom: 0;
+		font-size: $fs-small;
+	}
+}
+
```
JPrevost: dude wtf is up with this spacing? Very not-:rainbow:

Honestly though, I kinda don't care what is in the scss though so if you prefer whatever atrocity this is, I can adjust my styles editor to not freak out at this stuff.

frrrances: our WP repos are tabs so i had that set as default. now that i'm working in repos that not have that, i change.

PR #221, opened 2017-06-30
Initial system to store local search hints

JPrevost:
## Status
**READY**

#### What does this PR do?
This allows for phrase matching of search queries to suggested
  'best bets' to display in a placard system.

Users are having difficulty locating some items. As the search engine is vended, we cannot adjust the upstream settings directly (we are working with them to address immediate concerns). This system will allow us to intervene in such cases that the search engine predictably does not return expected results.

This is preliminary work currently behind a feature flag. Additional work will be done
before we make this a public facing feature.

- migrate your db to get the latest schema (see below)
- create a Hint using a rails console: `heroku local:run rails console`
- in the rails console:
  `Hint.create(title: 'yo title', url: 'http://example.com/get/stuff', fingerprint: 'popcorn')`
- enable the feature in your browser sessions: `/toggle?feature=hints`
- search for `popcorn` and see a placard!

- https://mitlibraries.atlassian.net/browse/DI-411

#### Todo:
- [x] Tests
- [ ] Documentation
- [ ] Stakeholder approval

YES
`bin/rails db:migrate` in your dev environment. This normally also updates your test db, but
if you run into issues, `bin/rails db:test:prepare` is your friend.

NO

### DIFF
--- a/app/controllers/hint_controller.rb
+++ b/app/controllers/hint_controller.rb
@@ -1,0 +1,7 @@
+class HintController < ApplicationController
+  def hint
+    return if params[:q].blank?
+    @hint = Hint.match(params[:q])
+    render layout: false
+  end
+end
--- a/app/models/application_record.rb
+++ b/app/models/application_record.rb
@@ -1,0 +1,3 @@
+class ApplicationRecord < ActiveRecord::Base
+  self.abstract_class = true
+end
--- a/app/models/hint.rb
+++ b/app/models/hint.rb
@@ -1,0 +1,5 @@
+class Hint < ApplicationRecord
+  def self.match(searchterm)
+    Hint.find_by(fingerprint: searchterm)
+  end
+end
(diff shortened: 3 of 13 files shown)


JPrevost:
The PR app has been populated with placards for `popcorn` and `indstat` searches for your demo-ing purposes.

thatandromeda:
I really appreciate how Hint.match() is separated out, so that it will be easy for me to fill in with more complex fingerprinting logic (and test).

I suspect the linktext wants to be different - like we want a 'best bet' type title for the hint box, and then use the hint.title rather than the hint.url as the linktext - but this doesn't have to block anything right now.

LGTM.

JPrevost:
Yeah, I want to work with UX ( aka @frrrances ) to revisit the mockup this is based on. I agree we probably want a "description" of some sort rather than displaying likely awful URLs. I'll merge so you can do your thing and we can iterate until it is 🌈 before un-feature flagging it.

PR #222, opened 2017-07-03
Add fingerprinting

thatandromeda:
## Status
**READY**

This will still be hidden behind a feature flag when deployed because we don't yet have the data to back it up, but the code itself is ready.

#### What does this PR do?
This adds fingerprinting to the `Hint.match()` logic, so that we can give users hints matching clusters of syntactically different but probably-semantically-identical search terms (e.g. "Prevost, Jeremy" and "jeremy prevost" are the same search). It uses the [OpenRefine fingerprinting algorithm](https://github.com/OpenRefine/OpenRefine/wiki/Clustering-In-Depth), with a slight modification to allow us to distinguish `C`, `C++`, and `C#` (the basic fingerprinting algorithm throws away all punctuation, but these C variants are common search terms at MIT and users expect them to be distinct).

There is an open Google doc containing our most common searches, waiting for user-services librarians to populate it with appropriate results so that we can seed the database. [Documentation on how I got those out of OpenRefine and our search logs.](https://github.mit.edu/pages/m31/2017/06/30/openrefine.html)

* `heroku local:run rails console`
* in the rails console:
  * `Hint.create(title: 'Web of Science', url: 'http://example.com/get/stuff', fingerprint: 'of science web')`
  * `Hint.create(title: 'Web of Science', url: 'http://example.com/get/stuff', fingerprint: 'knowledge of web')`
  * `Hint.create(title: 'Introduction to C', url: 'http://example.com/get/stuff', fingerprint: 'c introduction to')`
  * `Hint.create(title: 'Introduction to C++', url: 'http://example.com/get/stuff', fingerprint: 'c++ introduction to')`
* enable the feature in your browser sessions: /toggle?feature=hints
* search for "Web of Science" and see a placard!
* search for "Web of Knowledge" and see the same placard! (these are both really common searches and they should yield identical results - 'web of science' is the indexed name but 'web of knowledge' is the branding)
* search for "Introduction to C" (or "introduction to c") and see a placard!
* search for "Introduction to C++" (or "introduction to c++") and see a placard, but it's different from the last one!

- https://mitlibraries.atlassian.net/browse/DI-420

#### Todo:
- [x] Tests
- [ ] Documentation
- [ ] Stakeholder approval

NO

YES


thatandromeda:
I would especially like your review on the regex in `Hint.fingerprint`--  it's ugly and I'm not sure it does what I want. I wrote a bunch of tests and they pass, but if you have other ideas for evil edge cases, please do test them (or tell me what I should be testing).

thatandromeda:
Also there's this one test that fails ONLY SOMETIMES - it is failing in the build here - and I cannot see why. So it'd be awesome if you know.

app/models/hint.rb
```diff
@@ -1,5 +1,36 @@
...
+  def self.fingerprint(searchterm)
+    # Preprocess.
+    searchterm.strip!
+    searchterm.downcase!
+    # What's up with the weird regexp? We want to remove punctuation from
+    # the string EXCEPT where it where it is part of the term C++ or C#
+    # (as a standalone term - appearing either immediately after whitespace
+    # or at the beginning of a line). Those terms are meaningfully different
+    # from C, and our users frequently search for them.
+    searchterm.gsub!(/(?<!\s[cC]|\s[cC]\+|^[cC]|^[cC]\+)[[:punct:]]/, '')
```
JPrevost: I might move this regex to it's own private method as it's makes reading `self.fingerprint` harder to understand as it's difficult to ignore the regex inline like this. Instead calling `depunctuate` or something would probably make it easier to read and you'd only get caught up in the complicated bits if you actually needed to. Maybe. No big deal either way.

app/models/hint.rb
```diff
@@ -1,5 +1,36 @@
+require 'stringex/core_ext'
```
JPrevost: You might not need be required to `require` here (ha!) as most gems auto require themselves when they are loaded from the `Gemfile`

app/models/hint.rb
```diff
@@ -1,5 +1,36 @@
...
+
 class Hint < ApplicationRecord
   def self.match(searchterm)
-    Hint.find_by(fingerprint: searchterm)
+    searchprint = fingerprint(searchterm)
+    Hint.find_by(fingerprint: searchprint)
   end
+
+  # This implements the OpenRefine fingerprinting algorithm. See
+  # https://github.com/OpenRefine/OpenRefine/wiki/Clustering-In-Depth#fingerprint .
```
JPrevost: rubocop will stop yelling about this if you remove the trailing ` .` from the url. It'll still be too long, but it's
reasonable smart about not freaking out about stuff like this.

app/models/hint.rb
```diff
@@ -1,5 +1,36 @@
...
+    term_array.sort!
+
+    # Rejoin tokens.
+    fingerprint = term_array.join(' ')
+
+    # Normalize to ASCII (e.g. gödel and godel are liable to be intended to
+    # find the same thing). (Yes, to_ascii lacks a to_ascii! equivalent.)
+    fingerprint.to_ascii
+  end
+
```
JPrevost: I support rubocop yelling about this extra space ;)

app/models/hint.rb
```diff
@@ -1,5 +1,36 @@
...
+    term_array = searchterm.split(' ')
+
+    # Remove duplicates and sort.
+    term_array.uniq!
+    term_array.sort!
+
+    # Rejoin tokens.
+    fingerprint = term_array.join(' ')
+
+    # Normalize to ASCII (e.g. gödel and godel are liable to be intended to
```
JPrevost:
```
# rubocop:disable AsciiComments
# 这包含一些私人信息 ...
# ...
# rubocop:enable AsciiComments
```

app/models/hint.rb
thatandromeda: Can you clarify what you need by 'might not' and 'most'?

JPrevost: Might not: "I don't think it's required, but if you tried it without adding that and it didn't work I guess it is"

JPrevost: If you don't want to auto-require, you can `require: false` in Gemfile.

https://stackoverflow.com/questions/4800721/bundler-what-does-require-false-in-a-gemfile-mean

app/models/hint.rb
thatandromeda: Naturally this is a thing that is _preferred_ for Python, and I had therefore configured my editor to auto-add newlines at EOF. *figures out how to make that setting language-specific*

app/models/hint.rb
JPrevost: fwiw, I believe this gem requires that you require it. Fun!

JPrevost: Hmmm. I guess not. I just had some unlucky failed tests (I removed it locally then ran a few test runs and they all failed... but that was just a string of unlucky flakey tests. It works fine removed.)

app/models/hint.rb
```diff
@@ -1,5 +1,52 @@
...
-end
+
+  # This implements the OpenRefine fingerprinting algorithm. See
+  # https://github.com/OpenRefine/OpenRefine/wiki/Clustering-In-Depth#fingerprint
+  def self.fingerprint(searchterm)
+    # Preprocess.
+    searchterm.strip!
+    searchterm.downcase!
+
+    depunctuate!(searchterm)
```
JPrevost: thanks for moving this to it's own method. much easier to follow this method with the 🙈 moved to `depunctuate!` :)

test/models/hint_test.rb
```diff
@@ -37,4 +37,136 @@ class HintTest < ActiveSupport::TestCase
...
+                 Hint.fingerprint('Introduction to C++'))
+
+    assert_equal(Hint.fingerprint('Introdüction to C++'),
+                 Hint.fingerprint('Introduction to C++'))
+
+    assert_equal(Hint.fingerprint(' Introduction to C++' ),
+                 Hint.fingerprint('Introduction to C++'))
+
+    assert_equal(Hint.fingerprint('C++: an introduction'),
+                 Hint.fingerprint('C++ an introduction'))
```
JPrevost: holy 🌈 you made it work!

PR #225, opened 2017-07-12
Include libraries.mit.edu domain in relevant_links

JPrevost:
## Status
**READY**

#### What does this PR do?

a change to how we handle "get URLs" necessitates we include the
libraries.mit.edu domain in our `relevant_links`

In order to not use irrelevant links to t.o.c or publisher sites that we include in our MARC records, we only count links to "relevant" domains as links for our `View Online` links.

Once the data is loaded from our test barton catalog into the test EDS profile, searching for records with get urls should provide that URL directly as the "View Online" url rather than as a passed parameter in a SFX url.

- https://mitlibraries.atlassian.net/browse/DI-436

#### Todo:
- [x] Tests
- [x] Stakeholder approval

NO

NO

### DIFF
--- a/app/models/result.rb
+++ b/app/models/result.rb
@@ -43,7 +43,8 @@ class Result
 
   # domains we consider relevant when evaluating links
   def relevant_links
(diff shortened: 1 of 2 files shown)


thatandromeda:
👍

PR #226, opened 2017-07-20
Explicitly set cache expiration

JPrevost:
## Status
**READY**

#### What does this PR do?

this explicitly sets the cache of search results to expire

* memcachier is supposed to purge based on LRU, but we have seen in
  production that didn't happen with our current configuration
* the intent here is to set cached items to explicitly expire so
  hopefully they will be purged appropriately as expected
* we'll need to monitor the results of this change to ensure it is
  working as intended (the cache size on staging is much smaller so
  we can use that to confirm expected results before production cache
  size gets large enough to be a concern)

Sadly, you can't really :/

- https://mitlibraries.atlassian.net/browse/DI-383 (sort of)

NO

NO

### DIFF
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -31,7 +31,7 @@ class SearchController < ApplicationController
   # items when necessary.
   def search_results(page, per_page)
     return unless valid_target?
-    Rails.cache.fetch(cache_key(page, per_page)) do
+    Rails.cache.fetch(cache_key(page, per_page), expires_in: 1.day) do
       search_target(page, per_page)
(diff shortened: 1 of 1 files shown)


thatandromeda:
LGTM _but_: how will we be monitoring to ensure this has the desired result?

JPrevost:
@thatandromeda monitoring for the problem remains an issue. For now I'm planning on keeping the ticket open as it's not really "solved" it's just "hopefully solved". I ran into problems trying to run `memcached` locally reliably at all so replicating the problem then proving the solution is evading me at this time.

PR #227, opened 2017-07-20
Handles records missing expected metadata

JPrevost:
## Status
**READY**

#### What does this PR do?

This fixes a bug in which we failed to recover from records missing expected metadata

Search for `National Cyclopaedia of American Biography` in production and in this PR build. One works!

#### Todo:
- [x] Tests

NO

NO

### DIFF
--- a/app/models/normalize_eds_articles.rb
+++ b/app/models/normalize_eds_articles.rb
@@ -35,6 +35,7 @@ class NormalizeEdsArticles
   end
 
   def doi
+    return unless bibrecord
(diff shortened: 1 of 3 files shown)


thatandromeda:
Any day which adds exception handling is a day that makes the world a little bit better. 👍

JPrevost:
@thatandromeda I was way too lazy in the initial coding of this. I am generally much more defensive in handling data from APIs. I _think_ most of the laziness has been protected now.

PR #228, opened 2017-07-21
Loads Hints from web accessible marcxml

JPrevost:
## Status
**READY**

#### What does this PR do?

* loads hints from web accessible marcxml
* fingerprints are generated from title and alt title fields

After setting an appropriate value for `ALEPH_HINT_SOURCE` (see LastPass or Heroku directly) you can:

```
heroku local:run rails console
Hint.count (observe small number)
AlephHint.new.process_records
(wait)
Hint.count (observe much larger number)
```

- https://mitlibraries.atlassian.net/browse/DI-445

#### Todo:
- [x] Tests
- [x] Documentation
- [x] Stakeholder approval

NO

YES, but not a gem dependency. In this case, `ALEPH_HINT_SOURCE` has been added to ENV

### DIFF
--- a/app/models/aleph_hint.rb
+++ b/app/models/aleph_hint.rb
@@ -1,0 +1,81 @@
+# Parses Aleph marcxml and creates Hints
+#
+require 'open-uri'
+class AlephHint
+  def initialize
+    @marcxml = marcxml
+  end
+
+  # Load aleph hint source as XML
+  def marcxml
+    Nokogiri::XML(open(Rails.application.secrets.aleph_hint_source))
+  end
+
+  # Selects the record nodes from the XML
+  def records
+    @marcxml.xpath('//xmlns:record')
+  end
+
+  # Loops over records and calls out to hint creation logic
+  def process_records
+    records.each do |record|
+      record_to_hints(record)
+    end
+  end
+
+  # Creates Hints for each record for each title and alt title
+  # NOTE: at this time there is no detection of duplicates as that logic
+  # may better reside in the Hint object itself to handle duplicates that may
+  # originate from different loaders.
+  def record_to_hints(record)
+    Hint.create(title: title(record), url: best_url(record),
+                fingerprint: Hint.fingerprint(title(record)))
+
+    fingerprint_alt_titles(246, record)
+
+    fingerprint_alt_titles(740, record)
+  end
+
+  # Creates a Hint for each alternate title
+  def fingerprint_alt_titles(field, record)
+    alt_titles(field, record).each do |alt_title|
+      Hint.create(title: title(record), url: best_url(record),
+                  fingerprint: Hint.fingerprint(alt_title))
+    end
+  end
+
+  # Extracts title from marc 245
+  def title(record)
+    extract_marc_value(245, 'a', record).text
+  end
+
+  # Extracts array of alternate titles from supplied tag (246 and 740)
+  def alt_titles(tag, record)
+    extract_marc_value(tag, 'a', record).map(&:text)
+  end
+
+  # Generic marc tag / subfield parser
+  def extract_marc_value(tag, subfield, record)
+    record.xpath(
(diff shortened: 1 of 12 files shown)


thatandromeda:
OK. I have various questions which you can address or not address as you prefer. I _do_ want stronger exception handling around the `best_url` situation, including a test of what happens if we get a totally bogus record without a trustworthy URL.

Also we have to actually resolve what happens with duplicate fingerprints, but I think having the model handle that and thereby taking it up in a separate PR is okay - we'll Slack about that one.

thatandromeda:
👍 (tell me if it's going to require another review when you have merged/rebased master)

app/models/aleph_hint.rb
```diff
@@ -0,0 +1,62 @@
...
+  def initialize
+    @marcxml = marcxml
+  end
+
+  def marcxml
+    Nokogiri::XML(open(Rails.application.secrets.aleph_hint_source))
+  end
+
+  def records
+    @marcxml.xpath('//xmlns:record')
```
JPrevost: no

JPrevost: (and if it did we'd get an exception and alerted... I'm not sure what you are getting at)

app/models/aleph_hint.rb
```diff
@@ -0,0 +1,62 @@
...
+    record.xpath(
+      "./xmlns:datafield[@tag=#{tag}]/xmlns:subfield[@code=\"#{subfield}\"]"
+    )
+  end
+
+  def url(record)
+    if extract_marc_value(856, 'u', record).present?
+      extract_marc_value(856, 'u', record).text
+    else
+      extract_marc_value(956, 'u', record).text
```
JPrevost: Unless the datasource changes, which would involve a discussion we'd be involved in, we can assume all records have a URL. If that contract is broken, we throw exceptions because things are broken.

app/models/aleph_hint.rb
```diff
@@ -31,27 +36,33 @@ def record_to_hints(record)
...
 
+  # Generic marc tag / subfield parser
   def extract_marc_value(tag, subfield, record)
     record.xpath(
       "./xmlns:datafield[@tag=#{tag}]/xmlns:subfield[@code=\"#{subfield}\"]"
     )
   end
 
+  # For some reason, we sometimes use the standard 866$u for URLs but other
+  # times use 956$u. :facepunch:
```
JPrevost: noted. I concur and will adjust accordingly. thx.

app/models/aleph_hint.rb
```diff
@@ -0,0 +1,62 @@
...
+                fingerprint: Hint.fingerprint(title(record)))
+
+    fingerprint_alt_titles(246, record)
+
+    fingerprint_alt_titles(740, record)
+  end
+
+  def fingerprint_alt_titles(field, record)
+    alt_title(field, record).each do |hint|
+      Hint.create(title: title(record), url: url(record),
```
JPrevost: yeah, `hint` is the wrong name. I'll update.

app/models/aleph_hint.rb
```diff
@@ -64,10 +64,17 @@ def extract_marc_value(tag, subfield, record)
...
-    end
+    urls = []
+    urls << extract_marc_value(956, 'u', record).map(&:text)
+    urls << extract_marc_value(856, 'u', record).map(&:text)
+    urls.flatten.compact.uniq
+  end
+
+  # Select the first Get URL from the array of urls
+  # NOTE: the datasource was created by selecting records that contained
+  # this url pattern so we make the assumption here that at least one url
```
JPrevost: We discussed this and agreed throwing an exception (which is what will happen) was ok. A later commit added more clarification to this note. I'm happy to discuss again if you feel we didn't adequately resolve your concern.

JPrevost:
Thanks for the review. I intend to rename a variable and remove an inappropriate comment before merging.

PR #232, opened 2017-07-31
421 google data loader

thatandromeda:
## Status
READY

#### What does this PR do?
Adds data loader for our handcrafted fingerprint data.

This assumes people will be dumping the data to a Dropbox URL - this makes it at least theoretically possible to automate in future (I don't want to take it straight from Sheets as people demonstrably use that as a space for work-in-progress). I've added some instructions to our Google Sheet to cover the export situation.

In your rails console, `Hint.where(source: 'custom').count`. Lo, there are none!

`heroku local:run rails reloadhints:custom[https://www.dropbox.com/s/f6p4qtj3chaqlp2/test_popular.csv?dl=0]`

In your rails console again, `Hint.where(source: 'custom').count`. Lo, there are many!

- [DI-421](https://mitlibraries.atlassian.net/projects/DI/issues/DI-421)

#### Todo:
- [x] Tests
- [x] Documentation
- [ ] Stakeholder approval

NO

NO


app/models/custom_hint.rb
```diff
@@ -0,0 +1,87 @@
...
+      %w{Title URL Fingerprint}.all? { |header| mycsv.headers.include? header }
+  end
+
+  # Loop over records and create hints. Make sure to filter out blank data -
+  # hand-created records may not have filled in all desired data, e.g. if there
+  # is not an obvious result.
+  def process_records
+    validate_csv
+    CSV.parse(@csv, :headers => true) do |record|
+      if %w{Title URL Fingerprint}.none? { |header| record[header].nil? }
```
JPrevost: I'm confused as to what is going on this this line. I will dig in on a console to wrap my head around this (:coffee: needs to kick in a bit more too). I am not suggesting we change it but I wanted to make a note to myself to come back to this without blocking the rest of the review.

JPrevost: Okay, I think the double negatives are confusing to me. Consider maybe this alternative:

```
if %w{Title URL Fingerprint}.all? { |header| record[header].present? }
```

app/models/custom_hint.rb
```diff
@@ -0,0 +1,87 @@
...
+
+  def process_url
+    canonicalize_url
+    validate_url
+  end
+
+  def validate_url
+    as_uri = URI(@url)
+
+    raise "Invalid URL - not a Dropbox download URL" unless [
```
thatandromeda: Ensuring that it ends in `dl=1` is dropbox-specific; in general I can be reasonably confident that the thing I'm getting served is *a file* and not *an html page* (that I would have to parse in a different way) by doing this, whereas if I were allowing more general URLs I'd want to be more paranoid about validation.

app/models/custom_hint.rb
thatandromeda: I'm saying "make sure the CSV file has all of these headers", but I have no idea if the way I'm doing it is Quintessentially Rubyish or Needlessly Obscurantist. (In Python it would be a list comprehension, but obviously that's Python.)

README.md
```diff
@@ -25,6 +25,14 @@ heroku local:run bin/rails reloadhints:aleph
 ```
 _note_: `reloadhints:aleph` requires ENV `ALEPH_HINT_SOURCE`
 
+To reload custom hints in a development environment:
+```
+heroku local:run bin/rails reloadhints:custom[https://www.dropbox.com/blah/blah]
```
thatandromeda: My thought process for going with a parameter here was something like:
* There should be a specific URL which the loader always looks at!
* That requires that we have a multi-person-writeable Dropbox folder, covering all the people who might have control over the hinting process, and that people update the same file (overwriting it in the same location) rather than uploading a new file.
* Ooooh, I don't trust humans.
* This is more likely to work in practice if we get the URL.

The AlephHint loader comes out of a different workflow, where 1) the data source is automated and 2) we have control over the resultant URL (I *also* have no idea how Dropbox generates URLs, hence no reason to be confident they'd be stable). So I am sensitive to your concerns about the loaders presenting a different interface, but I also think it might be worth it in lieu of the fact that the underlying workflows are in fact quite different.

(Also I think ideally we automate all the loaders which *do* present stable interfaces and then we don't have to think about it any more.)

app/models/custom_hint.rb
```diff
@@ -0,0 +1,87 @@
...
+      as_uri.host == 'www.dropbox.com',   # ...from Dropbox
+      as_uri.query == 'dl=1'              # ...which triggers a file download
+    ].all?
+  end
+
+  # If a dropbox URL ends in ?dl=1, it downloads the file; if it ends in ?dl=0,
+  # it renders the file as HTML. We want to make sure we download the file, but
+  # the copy-link version gives the render-page option by default, so let's
+  # recover gracefully in that case.
+  def canonicalize_url
```
thatandromeda: Huh, can you expand on this? My preference is to gracefully recover from common problems when we know how rather than bothering the humans about it. (In this case, the default Dropbox share link is `dl=0` but the format we need is `dl=1`; we can tell people we want the `dl=1` form but the odds they will forget seem high to me, so I'd rather just DWIM.)

test/vcr_cassettes/custom_hint.yml
```diff
@@ -0,0 +1,276 @@
+---
+http_interactions:
+- request:
+    method: get
+    uri: https://www.dropbox.com/s/f6p4qtj3chaqlp2/test_popular.csv?dl=1
```
thatandromeda: Ahhhh, now I get what that line was doing.

Also, VCR is *amazing*.

app/models/custom_hint.rb
JPrevost: Yeah, I'm not sure if you saw my later comment. Please consider positive assertions rather than the double negative. I feel like it's easier to grok what is happening but results in the same logic.

app/models/custom_hint.rb
JPrevost: I'm still not sold on requiring dropbox URLs as the only loadable source of this content.

README.md
JPrevost: Fair enough.

app/models/custom_hint.rb
thatandromeda: I did not see the later one. Sure, works for me.

app/models/custom_hint.rb
```diff
@@ -0,0 +1,87 @@
+# Loads our handcrafted hints. We collaborate on these in a google sheet, but
+# we draw them from a separate CSV file rather than automatically from Sheets
+# for two reasons:
+#  1) to avoid having a production app have reliance on a developer's personal
```
thatandromeda: Done

app/models/custom_hint.rb
thatandromeda: OK, so if we let URLs be anything, what does validation look like? We need to handle the fact that an arbitrary URL might be serving up our content as HTML (as is the case with the AlephHints) or as a downloadable attachment. We need to make sure that we're following whatever redirect process might need to be followed, but also that we're not hanging if we find ourselves in an authentication process because people forgot to make the file world-readable. What else?

JPrevost: Your CSV parser would throw an invalid CSV error and rollback... right?

app/models/custom_hint.rb
JPrevost: I'm not sure if this is confusing because of the rebase, but it looks like this is still here in the final commit?

JPrevost:
👍

PR #237, opened 2017-08-08
Calculate fingerprints as needed when loading CustomHints

thatandromeda:
Users may have added new search terms to the custom spreadsheet -
they probably don't know how to calculate fingerprints, so we should do
that when fingerprints are absent.

## Status
**READY**

#### What does this PR do?
Adds fingerprints as needed when loading CustomHints.

While I generated the original custom hint spreadsheet via a data dump which included calculated fingerprints, in testing we decided there are additional custom terms we should hint on. Humans adding rows to the spreadsheet are unlikely to know how to calculate hints, so we should make sure to fill them in.

If you've got this branch on localhost in your vagrant,
`heroku local:run rails reloadhints:custom[https://www.dropbox.com/s/vm2xetxclk543zv/custom%20hint%20for%20test%20suite.csv?dl=1]`

Or the equivalent command on the PR build.

In the rails console, observe that `Hint.match('cnki')` works even though this example has no fingerprint in the loaded spreadsheet (which you can see at the dropbox URL above).

- https://mitlibraries.atlassian.net/browse/DI-471

#### Todo:
- [x] Tests
- [x] Documentation
- [ ] Stakeholder approval

NO

NO

### DIFF
--- a/app/models/custom_hint.rb
+++ b/app/models/custom_hint.rb
@@ -55,7 +55,8 @@ class CustomHint
   def validate_csv
     mycsv = CSV.new(@csv, headers: true).read
     raise 'Invalid CSV - wrong headers' unless \
-      %w(Title URL Fingerprint).all? { |header| mycsv.headers.include? header }
+      %w(Title URL Fingerprint Example\ search).all? \
+        { |header| mycsv.headers.include? header }
   end
 
   # Loop over records and create hints. Make sure to filter out blank data -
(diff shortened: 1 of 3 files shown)


app/models/custom_hint.rb
```diff
@@ -64,7 +65,13 @@ def validate_csv
   def process_records
     validate_csv
     CSV.parse(@csv, headers: true) do |record|
-      if %w(Title URL Fingerprint).none? { |header| record[header].nil? }
+      if %w(Title URL).all? { |header| record[header].present? }
+        # Add the fingerprint if needed; people creating spreadsheet rows may
```
JPrevost: Does there remain value in storing the fingerprint for _some_ records in the spreadsheet? Or should we always just ensure there is an example search and generate a fingerprint from that?

app/models/custom_hint.rb
```diff
@@ -64,7 +65,13 @@ def validate_csv
   def process_records
```
JPrevost: rubocop does flag this method as a bit long and a bit complex. I won't block merge on that as it's not egregious, but consider whether any of this logic could break out to another method to make it easier to read in the future.

app/models/custom_hint.rb
thatandromeda: I think you're right - I've been keeping them as a historical artifact (because they happened to be part of the original spreadsheet), but there's not logical value in keeping them going forward, and the code can certainly be simplified without them.

PR #238, opened 2017-08-09
451 prioritize by source

thatandromeda:
## Status
**READY**

#### What does this PR do?
Allows us to set an environment variable listing the Hint sources we display to the user, in order of priority.

We may have hints with the same fingerprint from multiple sources (e.g. one automatically generated from aleph and one manually entered by staff on the spreadsheet); we need a tiebreaker mechanism.

* In your console, create Hints that both have a fingerprint of `pubmed`, one whose source is `custom` and one whose source is `aleph`
* `Hint.match('pubmed')` and observe that you get the custom-sourced one
* Set `ENV['HINT_SOURCES'] = 'custom,aleph'`; note that `Hint.match('pubmed')` returns the custom-sourced hint
* Set `ENV['HINT_SOURCES'] = 'aleph,custom'`; note that `Hint.match('pubmed')` returns the aleph-sourced hint
* Set `ENV['HINT_SOURCES'] = 'fake'`; note that `Hint.match('pubmed')` returns `nil` as neither matching source is in your `HINT_SOURCES`

- https://mitlibraries.atlassian.net/browse/DI-452

#### Todo:
- [x] Tests
- [x] Documentation
- [ ] Stakeholder approval

NO

NO

### DIFF
--- a/app/models/hint.rb
+++ b/app/models/hint.rb
@@ -20,10 +20,6 @@ class Hint < ApplicationRecord
   validates :source, presence: true
   validates :fingerprint, uniqueness: { scope: :source }
 
-  # ~~TODO~~ testing:
-  # hints are matched in order of priority
-  # if env is unset, sources defaults to custom
-  # hints in db but not in HINT_SOURCES are not returned
   def self.match(searchterm)
     searchprint = fingerprint(searchterm)
     hints = Hint.where(fingerprint: searchprint)
@@ -96,6 +92,6 @@ class Hint < ApplicationRecord
     if ENV['HINT_SOURCES'].nil?
       return ['custom']
     end
-    ENV['HINT_SOURCES']
+    ENV['HINT_SOURCES'].split(',')
   end
 end
--- a/app/models/hint.rb
+++ b/app/models/hint.rb
@@ -25,7 +25,7 @@ class Hint < ApplicationRecord
     hints = Hint.where(fingerprint: searchprint)
     # Order matching hints by source; throw away nils; return first remaining
     # Hint. If there are none, this will return nil.
-    sources.map { |source| hints.find_by(source: source) }.compact[0]
+    sources.map { |source| hints.find_by(source: source) }.compact.first
   end
 
   # Updates a record if it exists, creates one if it does not
@@ -89,9 +89,6 @@ class Hint < ApplicationRecord
 
   def self.sources
     # Set a default value.
-    if ENV['HINT_SOURCES'].nil?
-      return ['custom']
-    end
-    ENV['HINT_SOURCES'].split(',')
+    ENV['HINT_SOURCES'].nil? ? ['custom'] : ENV['HINT_SOURCES'].split(',')
   end
 end
--- a/app/models/hint.rb
+++ b/app/models/hint.rb
@@ -20,9 +20,16 @@ class Hint < ApplicationRecord
   validates :source, presence: true
   validates :fingerprint, uniqueness: { scope: :source }
 
+  # ~~TODO~~ testing:
+  # hints are matched in order of priority
+  # if env is unset, sources defaults to custom
+  # hints in db but not in HINT_SOURCES are not returned
   def self.match(searchterm)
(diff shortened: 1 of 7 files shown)


test/models/hint_test.rb
```diff
@@ -26,6 +35,39 @@ class HintTest < ActiveSupport::TestCase
...
+    match = Hint.match('of science web')
+    assert_equal('custom', match.source)
+
+    hints(:webofscience).destroy
+    match = Hint.match('of science web')
+    assert_equal('aleph', match.source)
+  end
+
+  test 'Hint has default sources if env not set' do
+    ENV.delete('HINT_SOURCES')
```
JPrevost: 🌈 Sweet. I missed this simple way of removing an set ENV in testing. I have generally resorted to setting the value to none with is way-less-:rainbow:

app/models/hint.rb
```diff
@@ -22,7 +22,10 @@ class Hint < ApplicationRecord
 
   def self.match(searchterm)
     searchprint = fingerprint(searchterm)
-    Hint.find_by(fingerprint: searchprint)
+    hints = Hint.where(fingerprint: searchprint)
+    # Order matching hints by source; throw away nils; return first remaining
+    # Hint. If there are none, this will return nil.
+    sources.map { |source| hints.find_by(source: source) }.compact[0]
```
JPrevost: I find `sources.map { |source| hints.find_by(source: source) }.compact.first` a bit more readable. That said, it's clearly clear what is happening as is so feel free to ignore :)

app/models/hint.rb
```diff
@@ -83,4 +86,12 @@ def self.depunctuate!(searchterm)
     searchterm.gsub!(/GROSSHACKCPLUSPLUS/, 'c++')
     searchterm.gsub!(/GROSSHACKCSHARP/, 'c#')
   end
+
+  def self.sources
+    # Set a default value.
+    if ENV['HINT_SOURCES'].nil?
```
JPrevost: You can avoid an explicit return like:
```
    if ENV['HINT_SOURCES'].nil?
      ['custom']
    else
      ENV['HINT_SOURCES'].split(',')
    end
```

app/models/hint.rb
JPrevost: With large numbers of sources, this may get inefficient due to how ActiveRecord compiles the logic in this method. I am 100% _not_ suggesting a change as I don't think we should have lots of sources in production but since I was curious how it worked I did some testing. Basically, it ends up combining the queries and then doing once each due to the map... so for two sources we get:
```
[2] pry(main)> Hint.match('pubmed')
  Hint Load (0.2ms)  SELECT  "hints".* FROM "hints" WHERE "hints"."fingerprint" = ? AND "hints"."source" = ? LIMIT ?  [["fingerprint", "pubmed"], ["source", "custom"], ["LIMIT", 1]]
  Hint Load (0.1ms)  SELECT  "hints".* FROM "hints" WHERE "hints"."fingerprint" = ? AND "hints"."source" = ? LIMIT ?  [["fingerprint", "pubmed"], ["source", "aleph"], ["LIMIT", 1]]
=> #<Hint:0x007f8975d3fb48
 id: 90332,
 title: "PubMed",
 url: "http://libraries.mit.edu/get/pubmed",
 fingerprint: "pubmed",
 created_at: Wed, 09 Aug 2017 14:01:14 UTC +00:00,
 updated_at: Wed, 09 Aug 2017 14:01:14 UTC +00:00,
 source: "custom">
```
We have a combined index on those two fields so I think we are all good, just wanted to share the results of my curiosity.

thatandromeda: I share your feeling about the query, and chose not to care because these querysets will be so small that the performance impact will be negligible, whereas the benefits in terms of semantics and configurability are large. (Although I still feel a bit guilty about the inefficiency.)

JPrevost: If this becomes a problem (it won't!) we can rework this.

thatandromeda: Yup, in violent agreement.

PR #239, opened 2017-08-15
Sets Hints feature to enabled by default

JPrevost:
## Status
**READY**

#### What does this PR do?

We are leaving the feature flag in place, but defaulting to enabled to
allow us to easily turn it back off if this does not go over well. We
will remove the feature flag once this is stable in production.

I'll load the PR build with CustomHints. You should be able to search in the PR build for something that should return as a Hint and see it without manually enabling the feature.

- https://mitlibraries.atlassian.net/browse/DI-484

NO

NO

### DIFF
--- a/config/features.rb
+++ b/config/features.rb
@@ -16,6 +16,6 @@ Flipflop.configure do
     description: 'Enables button for EDS supplied non-subscribed SFX links'
 
   feature :hints,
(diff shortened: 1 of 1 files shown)


thatandromeda:
🌈 🦄 😸 🎆

PR #244, opened 2017-08-22
Handle bad EDS Session Tokens

JPrevost:
## Status
**READY**

#### What does this PR do?

* EDS sometimes does not accept session tokens they have created
* detect and log when that happens
* attempt to create a new session token, but only once, when detected
  to hopefully provide end users with a reliable experience even when
  our upstream source is being flakey
* throw an exception if the bad token is detected on two consecutive attempts during a single  
  request

Sadly, there is no great way without changing the code. However, this is how to change the code to see what should happen.

In `search_eds.rb` model, change line 70 from `return unless eds_session_invalid?(result)` to `return if @attempts == 1`. When running a local search, your logs should now show the detection of a bad token, a retry, and then your search should succeed (unless you somehow manage to do this at the same time as EDS is for real having problems in which case you'll get the exception message).

- https://mitlibraries.atlassian.net/browse/DI-493

#### Todo:
- [x] Tests
- [x] Documentation

NO

NO

### DIFF
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -16,8 +16,8 @@ class SearchController < ApplicationController
     page = params[:page] || 1
     per_page = params[:per_page] || ENV['PER_PAGE'] || 20
     @results = search_results(page, per_page)
-    @pageable_results = paginate_results(page, per_page)
     return redirect_to root_url unless @results
+    @pageable_results = paginate_results(page, per_page)
     render 'search_boxed'
   end
 
--- a/app/models/search_eds.rb
+++ b/app/models/search_eds.rb
@@ -17,11 +17,16 @@ class SearchEds
   RESULTS_PER_BOX = ENV['RESULTS_PER_BOX'] || 5
 
   def initialize
+    @attempts = 0
     @auth_token = uid_auth
     @results = {}
   end
 
+  # Run a search, including creating and destroying EDS sessions.
   def search(term, profile, facets, page = 1, per_page = RESULTS_PER_BOX)
+    # store passed parameters so we can retry later if we need to.
+    @query = { term: term, profile: profile, facets: facets,
+               page: page, per_page: per_page }
     return 'invalid credentials' unless @auth_token
     @session_key = create_session(profile) if @auth_token
     raw_results = search_filtered(term, facets, page, per_page)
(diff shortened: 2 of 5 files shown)


app/controllers/search_controller.rb
```diff
@@ -16,8 +16,8 @@ def search
     page = params[:page] || 1
     per_page = params[:per_page] || ENV['PER_PAGE'] || 20
     @results = search_results(page, per_page)
-    @pageable_results = paginate_results(page, per_page)
     return redirect_to root_url unless @results
+    @pageable_results = paginate_results(page, per_page)
```
thatandromeda: 🌮 That is definitely the correct order.

(that was supposed to be a thumbs-up but I prefer the typo)

app/models/search_eds.rb
```diff
@@ -17,11 +17,16 @@ class SearchEds
   RESULTS_PER_BOX = ENV['RESULTS_PER_BOX'] || 5
 
   def initialize
+    @attempts = 0
```
thatandromeda: Please convince me this is thread-safe.

thatandromeda:
If this is thread-safe, I think it works, although it took me a little while to trace through all the possible data flows and convince myself.

That said, I'm unclear whether rails _models_ are thread-safe (I gather that controllers are, but we're handling this in a model). If they're not, I wonder if that might also bear on our session validity bug.

Speaking of sessions, this code might create up to three of them (once each when `@attempts` = 0, 1, and 2, since session creation happens before the exception is raised). Is that bad? Do we need to destroy sessions at some point? Are there problems that might be caused by having a bunch of them hanging around?

On a more meta level, do we even want to try handling the sessions ourselves, or do we want to instead refactor to use the EDS gem here and see if that makes session-handling their problem?

JPrevost:
We destroy sessions as part of the `#search` method during the happy path. I'm not too concerned with cleaning up sessions they have already invalidated on the non-happy path.

The core of the session bug we are seeing in production is not a thread safety issue. It is replicable outside of our code by Ebsco engineers, albeit it's sporadic so they haven't actually fixed it and when they do it'll take a while to actually convince me it's fixed.

I do want to eventually refactor to using the gem. When doing that, we can consider whether we maintain single API sessions for each Rails session. There is both benefit and cost for doing that.

app/models/search_eds.rb
JPrevost: We instantiate a new object for each request so... I'm going with I think so? If we reused the search objects this would be awful for sure. But since we create a new object on each search I think we're good:
https://github.com/MITLibraries/bento/blob/master/app/controllers/search_controller.rb#L68

thatandromeda:
OK, thanks for the thoughts. Looks like you will need to rebase but then merge away.

PR #246, opened 2017-08-24
MIT IP detection and helper methods

JPrevost:
~~Currently on hold as I want to test this on heroku via a proxied request as I've seen weird behavior with `remote_ip` in the past in some proxy configs.~~ From what I can tell this has been fixed in Rails.

## Status
**READY**

#### What does this PR do?

Provides the current MIT IP ranges we consider on campus for purposes
of discovery.

Includes helper methods to detect if the current request is coming
from an MIT IP (aka `member`) or non-MIT IP (aka `guest`).

Access the PR build from an MIT IP and note access level full in the footer.

Access the PR build from a non-MIT IP (including guest wireless) and note access level guest in the footer.

- https://mitlibraries.atlassian.net/browse/DI-496

#### Todo:
- [x] Tests
- [x] Documentation

NO

YES


app/helpers/search_helper.rb
```diff
@@ -8,4 +8,69 @@ def view_more_text
...
+  def access_level
+    if guest?
+      'guest'
+    else
+      'full'
+    end
+  end
+
+  # Current request IP is outside our campus range
+  def guest?
```
thatandromeda: Does this function really provide any utility? `access_level` could just use `member?`.

app/views/layouts/_institute_footer.html.erb
```diff
@@ -12,6 +12,7 @@
       </div>
       <div class="license">Licensed under the <a href="https://creativecommons.org/licenses/by-nc/2.0/" class="license-cc">Creative Commons Attribution Non-Commercial License</a> unless otherwise noted. <a href="https://libraries.mit.edu/research-support/notices/copyright-notify/">Notify us about copyright concerns</a>.
       </div><!-- end .footer-info-institute -->
+      <div>Access level: <%= access_level %></div>
```
thatandromeda: Why is this here? What's the rationale for displaying it to users, and has UX seen this?

thatandromeda:
You may tell codeclimate that you `wontfix` that issue 😉 .

Please do think about my questions, but I won't block you on them as I'm going to be on vacation in a few hours - address them however you think best.

app/views/layouts/_institute_footer.html.erb
JPrevost: Yeah, this may get removed by UX. I won't be merging until I've confirmed with them. I warned @frrrances about it but haven't actually confirmed they do or do not want this in there.

app/helpers/search_helper.rb
JPrevost: This was really built for use in the EDS API calls to send along with the API call. I think `guest?` will make that call clearer than `!member?`. `access_level` was a last minute addition and is possibly of no use at all.

JPrevost:
@frrrances can you weigh in on the Access Level display in the footer? I'm cool with removing it if we don't think it'll be useful information. Thx!

app/views/layouts/_institute_footer.html.erb
frrrances: I think it could be helpful, but for now it may make more sense in the source code. If we find that AskUs or librarians might need this to help debug access issues, we can talk about bringing it out to be more visible.

JPrevost: Cool. Thx.

PR #248, opened 2017-08-30
Records existance of hint in google analytics via staccato gem

matt-bernhardt:
## Status
**READY**

#### What does this PR do?
This improves our analytics capabilities around hints. Specifically, it:
- records interactions with hints in their own action, instead of alongside other branding elements.
- uses the staccato gem to record the _display_ of hints, in addition to just when users click on them.

- Inspect the markup around a displayed hint to see a data attribute for the container.
- Watch the real time events panel in Google Analytics to see server-side analytics when they are displayed.
- Watch the real time events panel in Google Analytics to see how interactions with hints are now re-classified in the `Bento / Hints` category / action, rather than `Bento / Link`

- https://mitlibraries.atlassian.net/browse/DI-474

#### Todo:
- [ ] Tests
- [ ] Documentation
- [ ] Stakeholder approval

NO

YES

### DIFF
--- a/app/controllers/hint_controller.rb
+++ b/app/controllers/hint_controller.rb
@@ -2,6 +2,10 @@ class HintController < ApplicationController
   def hint
     return if params[:q].blank?
     @hint = Hint.match(params[:q])
+    if (@hint && ENV['GOOGLE_ANALYTICS'])
+      tracker = Staccato.tracker(ENV['GOOGLE_ANALYTICS'], nil, ssl: true)
+      tracker.event(category: 'Bento', action: 'Hint', label: 'Show', value: 1)
+    end
     render layout: false
   end
 end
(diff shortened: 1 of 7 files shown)


JPrevost:
@matt-bernhardt Can you address the codeclimate warning?

This is the approach I would have taken with this. So 👍 overall. 🍻

app/controllers/hint_controller.rb
```diff
@@ -2,6 +2,10 @@ class HintController < ApplicationController
   def hint
     return if params[:q].blank?
     @hint = Hint.match(params[:q])
+    if (@hint && ENV['GOOGLE_ANALYTICS'])
```
JPrevost: I'll get back to you shortly on what we might care about in terms of testing for this block.

matt-bernhardt:
I appreciate being able to merge, but this should be easy enough to rebase (or just re-do). I'd prefer to understand the testing approach for this first; shortcutting the process isn't necessary since this is anticipatory rather than bug fix work.

JPrevost:
@matt-bernhardt I've added a feature flag which can be toggled via an ENV setting (so we can have it default to 'ON' in staging / PR builds but 'OFF' in production for a bit).

If this seems reasonable to you, you can merge once tests pass.

PR #254, opened 2017-09-08
514 fulltext link

thatandromeda:
## Status
**READY**

~~Opening the PR to get the PR build, but I want UX to spend some quality time evaluating the EBSCO fulltext links.~~ UX feedback acquired; this is ready to roll.

#### What does this PR do?
Adds a link to online fulltext, where available, using the EBSCO gem's logic for prioritizing fulltext sources.

Refactors `record.html.erb` and `RecordController` for readability.

* Go to the PR build
* Make sure to `/toggle/?feature=local_full_record`
* Search for stuff
* Look at its detail pages
* Lo, a 'view online' button where available
* Or a 'check for online copy' link if the top-choice link is SFX (see, e.g., https://mit-bento-staging-pr-254.herokuapp.com/record/edshtl/mdp.39015061775196)

- https://mitlibraries.atlassian.net/browse/DI-514

#### Todo:
- [ ] Tests
- [ ] Documentation
- [x] Stakeholder approval

NO

NO


thatandromeda:
Questions that have arisen from testing (I want to let these percolate for a while before deciding whether they are tickets, so I'm keeping track of them here):
* Should the fulltext link here be the same as the one on the search results page - if so, which logic do we use?
  * Note that we're using an EBSCO `Record` object on the full record page and a local `Result` object on the search page so there are exciting costs to switching either way
* Should we prioritize OA options if available?
  * Alternately, will they be separately indicated?

thatandromeda:
@JPrevost I need to add the tests but this is ready for your review now.

app/controllers/record_controller.rb
```diff
@@ -4,20 +4,42 @@ class RecordController < ApplicationController
...
+  def valid_url?
+    @record_source = params[:db_source]
+    @record_an = params[:an]
+    @record_source.present? && @record_an.present?
+  end
+
+  # If the top EDS fulltext link is an SFX link, we want to present an
+  # unstyled link to check for fulltext online. Otherwise we want an
+  # attention-getting 'view online' button
+  def use_sfx?
```
JPrevost: This may be better as a helper method and called from the view directly.

test/integration/record_test.rb
```diff
@@ -61,5 +61,19 @@ class RecordTest < ActionDispatch::IntegrationTest
...
     end
+
+  test 'view online button is shown' do
+    VCR.use_cassette('record: article', allow_playback_repeats: true) do
+      get record_url, params: { db_source: 'aci', an: '123877356' }
+      assert_select 'a.button-primary', text: 'View online'
+    end
+  end
+
+  test 'check for online copy link shown when fulltext URL is SFX' do
```
JPrevost: The EDS API provides a mechanism to know whether it is likely full text or not. That may not have made it to the Gem. If it did not, we probably want to open a feature request as many SFX links are known to be full text. You can't tell by the URL pattern, but EDS knows due to 🔮 (or more likely it knows because we told them our holdings).

app/helpers/record_helper.rb
```diff
@@ -57,4 +63,18 @@ def aleph_record?
       false
     end
   end
+
+  # If the top EDS fulltext link is an SFX link that does NOT go directly to a
+  # fulltext resource, we want to present an unstyled link to check for
+  # fulltext online. Otherwise, we want an attention-getting 'view online'
+  # button.
+  def check_online?
+    url = @record.fulltext_link[:url]
```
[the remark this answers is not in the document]
thatandromeda: The contents of `@record.fulltext_link[:label]` don't match the `Name` string in this example.  (In this record, we see `:label=>"Check SFX for availability"`; in another test record that *is* available online, we see `:label=>""`.)

One thing I've found that might work is the `eds_fulltext_word_count` property (which is 0 for the test case where it is not available online and nonzero for my test case where it is), but I definitely feel like I'm using an unsupported feature if I go this way. Do you know where to find any documentation that addresses this question? (I've looked at the github wiki and at edswiki without luck.) Or can you introduce me to the engineers you know at EBSCO so I can ask?

thatandromeda:
Adding a comment here to capture the upshot of a discussion in slack - we also need to look into pdflink handling in case they're expiring timebombs.

PR #262, opened 2017-09-22
497 login prompt

JPrevost:
## Status
**READY**

#### What does this PR do?

Activates guest mode for full record views.

Provides links to proxy site in production and override guest mode in development/test.

From a non-MIT IP access restricted records and records with expiring direct links.

https://mit-bento-staging-pr-262.herokuapp.com/record/mdc/25750248

From a non-MIT IP, instead of a direct PDFLINK, you'll see a link to sign in. Once signed in, you'll get the direct PDFLINK.

Any other redacted information should automatically be suppressed by the guest mode included in this PR (but I don't have examples of other content that is only accessible to signed in users).

- https://mitlibraries.atlassian.net/browse/DI-497

#### Screenshots

Guest mode
screen shot 2017-09-22 at 2 56 37 pm

non-Guest mode
screen shot 2017-09-22 at 2 56 42 pm

#### Todo:
- [ ] Tests
- [ ] Documentation
- [ ] Stakeholder approval

NO

NO

### DIFF
--- a/app/helpers/login_helper.rb
+++ b/app/helpers/login_helper.rb
@@ -1,0 +1,11 @@
+module LoginHelper
+  # Authentication Flow and Guest Mode documentation is available
+  # https://mitlibraries.atlassian.net/wiki/x/CoAeAw
+  def login_url
+    if !Rails.env.production?
+      request.original_url + '?guest=false'
+    else
+      ENV['PROXY_PREFIX'] + request.original_url
+    end
+  end
+end
--- a/app/views/record/_availability.html.erb
+++ b/app/views/record/_availability.html.erb
@@ -2,6 +2,8 @@
 <% if @record.fulltext_link.present? %>
   <% if check_online? %>
     <a href="<%= @record.fulltext_link[:url] %>">Check for online copy</a>
+  <% elsif guest_and_restricted_link? %>
+    <a href="<%= login_url %>">Sign in for full text link</a>
   <% else %>
     <a href="<%= @record.fulltext_link[:url] %>" class="btn button-primary">View online</a>
   <% end %>
--- a/README.md
+++ b/README.md
@@ -16,6 +16,10 @@ Search API. Appropriate credentials for both are required (see below).
 ## Bento System Overview
 ![alt text](docs/charts/bento_overview.png "Bento system overview chart")
 
+## Authentication Flow and Guest Mode Details
+
+https://mitlibraries.atlassian.net/wiki/x/CoAeAw
+
 ## Loading Hints
 
 Tasks exist to reload hints from supported sources.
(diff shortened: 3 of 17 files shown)


JPrevost:
The uncovered link is really not easily tested due to it requiring manipulation of the Rails.env. It's not impossible, it's just more work than is probably worthwhile.

JPrevost:
@frrrances you can see what I've come up with for login links here. Future PRs welcome ;)

JPrevost:
@thatandromeda this work flow requires documentation. I'm working on that. If you could review the work prior to the docs being complete, I'd appreciate it. If you find yourself unable to review due to documentation lacking though, please let me know and I'll re-prioritize that.

app/views/record/_availability.html.erb
```diff
@@ -2,6 +2,8 @@
 <% if @record.fulltext_link.present? %>
   <% if check_online? %>
```
thatandromeda: Thinking through what happens here...if they're not authenticated, when they click through to SFX, they're going to face an authentication check in there if required, right?

Is that what we want? Or do we want to check if they're guests *first* and always display "sign in for full text link", unless it's an OA resource?

(Did we just secretly solve the OA badging question?)

thatandromeda: Maybe what I am asking here is, has UX seen this logic?

JPrevost:
I'm confident now the in process documentation is indeed going to solve your questions. In short, login is a user friendly way of saying "use ezproxy". If they are using ezproxy, they are going to get an MIT IP because that's essentially how ezproxy works (all traffic to our app that is flowing through ezproxy will look like an MIT IP to us... so they will no longer be guests as long as they are being proxied).

app/views/record/_availability.html.erb
JPrevost: The added logic I put in here involves the time bombed pdf links. We _can't_ show those to people that are guests as they always return the full content with no security checks on the other end (argh!). This looks for signs of a pdflink and won't show it to guests.

app/helpers/login_helper.rb
```diff
@@ -0,0 +1,9 @@
+module LoginHelper
+  def login_url
```
JPrevost: Here's the `login_url` you were asking about

JPrevost:
@thatandromeda link to promised documentation. If this does not help, let's chat on slack to make sure the documentation sets the context necessary to understand how this works.

https://mitlibraries.atlassian.net/wiki/spaces/DI/pages/52330506/Authentication+Flow+and+Guest+Mode

thatandromeda:
Great, that is very helpful. Can you add a link to the documentation where future-us can find it in context (probably in `record_helper`)? Then go ahead and merge.

JPrevost:
Good call on adding a link to the docs where we may actually see it when we need it. Thanks for the review (of both the code and the docs).

PR #272, opened 2017-10-13
Adds expiring URL support

JPrevost:
## Status
**READY/**

#### What does this PR do?

EDS API sometimes returns expiring URLs. To allow users to have a
pleasant experience, we suppress display of those URLs until the users
clicks the "View Online" button at which point we ask for a _new_ copy
of the expiring URL and then redirect the user to that non-expired
version.

Also includes support to ensure guests are not taken to expiring URLs
as they are not authenticated on the target system.

On staging, go to https://mit-bento-staging.herokuapp.com/record/mdc/25750248

Note the View Online is an actual direct link (it will eventually expire and you'd be sad)

On the pr build, the same record https://mit-bento-staging-pr-272.herokuapp.com/record/mdc/25750248

Note the View Online link is a local route that will fetch a new copy of the EDS data and then redirect you to the intended URL with no fear of being sad due to expiring links.

- https://mitlibraries.atlassian.net/browse/DI-538

#### Todo:
- [x] Tests

NO

NO

### DIFF
--- a/app/controllers/record_controller.rb
+++ b/app/controllers/record_controller.rb
@@ -35,9 +35,10 @@ class RecordController < ApplicationController
   end
 
   def restricted!
+    return unless helpers.guest?
     flash[:error] = 'Restricted access. Please use our feedback form for ' \
                     'assistance.'
-    return redirect_to root_url if helpers.guest?
+    redirect_to root_url
   end
 
   def valid_url?
--- a/app/controllers/record_controller.rb
+++ b/app/controllers/record_controller.rb
@@ -1,4 +1,7 @@
 class RecordController < ApplicationController
+  before_action :restricted!, only: [:direct_link]
+  before_action :valid_url!
+
   rescue_from EBSCO::EDS::BadRequest, with: proc {
     raise RecordController::NoSuchRecordError, 'Record not found'
   }
@@ -8,8 +11,6 @@ class RecordController < ApplicationController
   include Rainbows
 
   def record
-    return redirect_to root_url unless valid_url?
-
     fetch_eds_record
     @keywords = extract_eds_text(@record.eds_author_supplied_keywords)
     @subjects = extract_eds_text(@record.eds_subjects)
@@ -20,6 +21,23 @@ class RecordController < ApplicationController
     @previous = params[:previous]
     rainbowify? if Flipflop.enabled?(:pride)
     render 'record'
+  end
+
+  def direct_link
+    fetch_eds_record
+    redirect_to @record.fulltext_link[:url]
+  end
+
+  private
+
+  def valid_url!
+    return redirect_to root_url unless valid_url?
+  end
+
+  def restricted!
+    flash[:error] = 'Restricted access. Please use our feedback form for ' \
+                    'assistance.'
+    return redirect_to root_url if helpers.guest?
   end
 
   def valid_url?
--- a/app/helpers/record_helper.rb
+++ b/app/helpers/record_helper.rb
@@ -119,8 +119,13 @@ module RecordHelper
     sanitize Nokogiri::HTML.fragment(CGI.unescapeHTML(input)).to_s
   end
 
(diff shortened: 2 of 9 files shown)


app/controllers/record_controller.rb
```diff
@@ -22,6 +23,23 @@ def record
...
   end
 
+  def direct_link
+    fetch_eds_record
+    redirect_to @record.fulltext_link[:url]
+  end
+
+  private
+
+  def valid_url!
```
thatandromeda: Is this stylistically better than the `return unless` construction, or is it just better in this context because it matches with the other `before_action`?

app/controllers/record_controller.rb
```diff
@@ -22,6 +23,23 @@ def record
...
+    return redirect_to root_url unless valid_url?
+  end
+
+  def restricted!
+    flash[:error] = 'Restricted access. Please use our feedback form for ' \
```
thatandromeda: This is a message we only want to show to guests, right? And in that case it would want to be inside the `if helpers.guest?` test. But as-is, wouldn't it actually display to everybody?

app/helpers/record_helper.rb
```diff
@@ -119,8 +119,13 @@ def safe_output(input)
     sanitize Nokogiri::HTML.fragment(CGI.unescapeHTML(input)).to_s
   end
 
+  # link is a direct expiring pdflink
+  def restricted_link?
+    @record.fulltext_link[:expires] == true
```
thatandromeda: Oooh, 🌈 that EBSCO provided us with this.

app/views/record/_availability.html.erb
```diff
@@ -1,15 +1,26 @@
 <h2 class="subtitle2">Availability</h2>
 <% if @record.fulltext_link.present? %>
   <div class="link-fulltext">
+
+    <%# SFX links that we don't know for sure if we own; auth happens at SFX %>
```
thatandromeda: This is starting to feel like a lot of logic for a view. I don't have a better idea, mind you, but I feel twitchy.

app/views/record/_availability.html.erb
```diff
@@ -1,15 +1,26 @@
...
         <p class="message">Full text available</p>
         <div class="actions">
           <a class="button button-secondary" href="<%= login_url %>">Sign in for access</a>
         </div>
       </div>
+
+    <%# Restricted expiring link, but current user is allowed to access it.
+        We'll get a fresh version of the link and redirect them to it %>
+    <% elsif restricted_link? %>
+      <%= link_to('View online', record_direct_link_path(params[:db_source], params[:an]), class: 'button button-primary green') %>
```
thatandromeda: 👍 Looks elegant, together with the controller logic.

app/controllers/record_controller.rb
JPrevost: Using before filters, we are running the `valid_url!` check for every route in this controller (because we don't declare which routes to run it on). The previous `return unless` was fine as it was only needed in one route, but once I added a second route, DRYing it up seemed better.

app/controllers/record_controller.rb
JPrevost: I concur this feels weird as is. I think it ends up working just fine, because either the user is fine and they end up redirected to the target url, or they are not fine and redirect to the root with the flash message. However, it would be better to _not_ set the flash unless they are a guest accessing a restricted link so I'll tweak to remove confusion.

app/controllers/record_controller.rb
thatandromeda: 👍

app/controllers/record_controller.rb
thatandromeda: 👍

JPrevost:
@thatandromeda I've added a commit to (hopefully) address your concern.

thatandromeda:
Go forth and mergeify.

PR #275, opened 2017-10-19
Frances/full record ui

frrrances:
## Status
**READY**

#### What does this PR do?
This PR adds some styling to the full record so that it matches the expected design.

- https://mitlibraries.atlassian.net/browse/DI-552

#### Screenshots (if appropriate)
search mit libraries

NO

NO

### DIFF
--- a/app/models/button_maker.rb
+++ b/app/models/button_maker.rb
@@ -19,7 +19,7 @@ class ButtonMaker
   def initialize(item, oclc)
     @item = item
     @oclc = oclc
-    @options = %w(call contact hold ill recall scan special_ill)
+    @options = %w(call contact hold recall ill scan special_ill)
 
     # Properties of items. This must go *after* setting @item and @oclc, but
     # *before* setting @requestable.
@@ -190,7 +190,7 @@ class ButtonMaker
 
   def make_button_for_ill
     "<a class='btn button-secondary button-small' "\
-      "href='#{url_for_ill}'>Get it with ILL (3-4 days)</a>"
+      "href='#{url_for_ill}'>Get non-MIT copy (3-4 days)</a>"
   end
 
   def make_button_for_recall
@@ -200,7 +200,8 @@ class ButtonMaker
   end
 
   def make_button_for_scan
-    "<a class='btn button-secondary' href='#{url_for_scan}'>Request Scan</a>"
+    "<a class='btn button-secondary button-small' " \
+      " href='#{url_for_scan}'>Request Scan</a>"
   end
 
   def make_button_for_special_ill
--- a/app/models/button_maker.rb
+++ b/app/models/button_maker.rb
@@ -19,6 +19,7 @@ class ButtonMaker
   def initialize(item, oclc)
     @item = item
     @oclc = oclc
+    # The order of this list controls the order in which buttons will display.
     @options = %w(call contact hold recall ill scan special_ill)
 
     # Properties of items. This must go *after* setting @item and @oclc, but
--- a/app/views/record/_basic_info.html.erb
+++ b/app/views/record/_basic_info.html.erb
@@ -30,8 +30,8 @@
   </span>
 <%# Includes Books; is catchall for other types not specified in wireframes %>
 <% else %>
-  <span class="record-type"><span class="sr">Type: </span> 
-    <%= @record.eds_publication_type %> 
+  <span class="record-type"><span class="sr">Type: </span>
+    <%= @record.eds_publication_type %>
   </span>
   <span class="record-year">
     Published <%= @record.eds_publication_year %>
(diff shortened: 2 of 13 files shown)


frrrances:
@JPrevost - one of the test failures is from the TODO note, is it possible to switch that test to "warn" instead of "fail" since the intent of the TODO note is to be able to catch that to fix later?

app/models/button_maker.rb
```diff
@@ -19,7 +19,7 @@ class ButtonMaker
   def initialize(item, oclc)
     @item = item
     @oclc = oclc
-    @options = %w(call contact hold ill recall scan special_ill)
+    @options = %w(call contact hold recall ill scan special_ill)
```
JPrevost: Does this control the order / priority of the buttons?

frrrances: yes! TIL :)

thatandromeda: Yup, and I'll add a comment accordingly.

app/views/record/_basic_info.html.erb
```diff
@@ -5,29 +5,49 @@
...
-  <% end %>
-<% end %>
+<p class="record-authors">
+  <span class="sr">Authors: </span>
+    <% @record.eds_authors.each do |author| %>
+    <span class="record-author">
+      <% if author == "et al" %>
+        <%= author %>
+      <% else %>
+        <%= link_to(author, search_prefix + "AU+\"#{author}\"", data: {type: "Author"} ) unless author == "et al" %><% if author != @record.eds_authors.last %><% end %>
```
thatandromeda: This line removes the semicolon separating authors - I have put it back in to match the wireframes; tell me if that's not what you want.

frrrances: i put it in with css

app/views/record/_basic_info.html.erb
```diff
@@ -5,29 +5,49 @@
...
     blank, corresponding to the wireframe. %>
-<% @record.eds_authors.each do |author| %>
-  <% if author == "et al" %>
-    <%= author %>
-  <% else %>
-    <%= link_to(author, search_prefix + "AU+\"#{author}\"", data: {type: "Author"} ) unless author == "et al" %><% if author != @record.eds_authors.last %>; <% end %>
-  <% end %>
-<% end %>
+<p class="record-authors">
+  <span class="sr">Authors: </span>
```
thatandromeda: This class will be apparent to people using screenreaders even if there are no authors (people reading the page with their eyes will simply not see an author section at all if there are no authors). What's the desired behavior here?

app/views/record/_extended_info.html.erb
```diff
@@ -4,91 +4,83 @@
...
-        <% clean_affiliations.each do |aff| %>
-          <li>
-            <%= aff %>
-          </li>
-        <% end %>
-      </ul>
-    </li>
+      <span class="label">Affiliations:</span> 
+      <% clean_affiliations.each do |aff| %>
+        <%= aff %>;
```
thatandromeda: This will have a semicolon after every affiliation, including the last one in the list; do you want to omit the semicolon after the last one?

Also "Affiliations" will be plural even if there is only one affiliation - do you want us to be clever about that?

app/views/record/_extended_info.html.erb
```diff
@@ -4,91 +4,83 @@
...
-    </li>
+      <span class="label">Publication info:</span> <%= safe_output(@record.eds_publication_info) %>
+    </li> 
   <% end %>
 
   <% if @record.eds_isbn_print.present? %>
     <li>
-      ISBN: <%= @record.eds_isbn_print %>
-    </li>
+      <span class="label">ISBN:</span> <%= @record.eds_isbn_print %>
```
thatandromeda: FYI in another branch I list all the ISBNs (and ISSNs). How about we let this branch merge *first*; then I'll rebase the other branch onto this one and you can tidy up the aesthetics as needed.

app/views/record/_sidebar.html.erb
```diff
@@ -1,12 +1,14 @@
-<h4>Sidebar</h4>
+<div class="bit">
+  <h3 class="title">Sidebar</h4>
 
-<ul>
-  <li>Advanced Search</li>
-  <li>Save to...</li>
+  <ul>
+    <li>Advanced Search</li>
```
thatandromeda: Do we know what these functions actually do yet? And if so, are there tickets for making them do that?

app/views/record/_basic_info.html.erb
frrrances: It would be nice if this was wrapped in an `if author(s)` statement.

app/views/record/_extended_info.html.erb
frrrances: I like those clever ideas. ++

app/views/record/_extended_info.html.erb
frrrances: Deal.

app/views/record/_sidebar.html.erb
frrrances: Oh yes, I suppose I could adjust this to be at least closer to real... brb.

PR #276, opened 2017-10-19
Handle Generic EDS errors

JPrevost:
## Status
**READY**

#### What does this PR do?

This detects and throws a specific exception when we detect EDS has returned an error.

Before this code, we through a generic exception that was not clear what was happening.

End user experience will remain the same as we handle the exception state in the UI in either case the same way.

- https://mitlibraries.atlassian.net/browse/DI-567

NO

NO

### DIFF
--- a/app/models/search_eds.rb
+++ b/app/models/search_eds.rb
@@ -58,25 +58,39 @@ class SearchEds
                                    connect: http_timeout,
                                    read: http_timeout)
                  .get(search_url(term, facets, page, per_page).to_s).to_s
-    detect_and_recover_from_bad_eds_session(result)
-    JSON.parse(result)
+    json_result = JSON.parse(result)
+    detect_eds_errors(json_result)
+    json_result
+  end
+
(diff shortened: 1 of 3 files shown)


JPrevost:
@thatandromeda I suggest we "it's fine for now" the class has 2 too many lines alert from CodeClimate.

app/models/search_eds.rb
```diff
@@ -58,25 +58,39 @@ def search_filtered(term, facets, page, per_page)
...
+  def detect_and_recover_from_bad_eds_session(json_result)
     prevent_multiple_retries
-    return unless eds_session_invalid?(result)
+    return unless eds_session_invalid?(json_result)
     Rails.logger.warn('EDS API Session Token Invalid')
     retry_query
   end
 
+  # Detect general EDS errors. Bad sessions still get special treatment in a
+  # seprate method to allow us to try to recover. Detecting bad sessions must
```
thatandromeda: 👍 useful comment

PR #279, opened 2017-10-20
remove extra colon, fixup sidebar

frrrances:
## Status
**READY**

#### What does this PR do?
A few small fixes from the last PR that didn't make the cut.

### DIFF
--- a/app/views/record/_basic_info.html.erb
+++ b/app/views/record/_basic_info.html.erb
@@ -47,7 +47,7 @@
         <% if author == "et al" %>
           <%= author %>
         <% else %>
-          <%= link_to(author, search_prefix + "AU+\"#{author}\"", data: {type: "Author"} ) %><% if author != @record.eds_authors.last %><% end %>
+          <%= link_to(author, search_prefix + "AU+\"#{author}\"", data: {type: "Author"} ) %> <%# semicolons between authors added by css %>
         <% end %>
       </span>
     <% end %>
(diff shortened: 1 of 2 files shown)


app/views/record/_basic_info.html.erb
```diff
@@ -47,7 +47,7 @@
         <% if author == "et al" %>
           <%= author %>
         <% else %>
-          <%= link_to(author, search_prefix + "AU+\"#{author}\"", data: {type: "Author"} ) %><% if author != @record.eds_authors.last %>;<% end %>
+          <%= link_to(author, search_prefix + "AU+\"#{author}\"", data: {type: "Author"} ) %><% if author != @record.eds_authors.last %><% end %>
```
thatandromeda: This removes all the colors (so if there are multiple authors, they come across as "Botsford, Frances Yelton, Andromeda" rather than "Botsford, Frances; Yelton, Andromeda"). Is there an example URL where you're seeing colons you don't expect?

frrrances: Hmm I was seeing double colons on the full record since I put them in using css and they were hardcoded here. Are you not seeing any?

thatandromeda: Ohhhh, OK. In that case I'm going to remove this if statement (it no longer does anything) and add a comment so people know to find semicolon-adding functionality in css.

thatandromeda:
I rebased this onto the new master - go ahead and merge once the tests pass.

PR #281, opened 2017-10-24
Force http thumbnail images to use https

JPrevost:
## Status
**READY**

#### What does this PR do?

Forces http thumbnails to use https (argh!)

If the url returned from the API is http, it changes it to https. If it is not http, we just use whatever we were given (either https already or something weird. Who knows!)

EDS API returns http thumbnail images. It appears the service they use can handle https. Some browsers are starting to get a bit more vocal with mixed content messages, and there is a desire for this application to avoid that if possible.

In the PR build, do a search and note the images we display in bento view and full record view are using https scheme. The same search in staging will show http scheme.

- https://mitlibraries.atlassian.net/browse/DI-509

#### Todo:
- [x] Tests
- [ ] Documentation
- [x] Stakeholder approval

NO

NO

### DIFF
--- a/app/helpers/record_helper.rb
+++ b/app/helpers/record_helper.rb
@@ -128,4 +128,21 @@ module RecordHelper
   def guest_and_restricted_link?
     guest? && restricted_link?
   end
+
+  # force http images to use https. If remote server cannot handle https a
+  # broken image is expected. The `alt` keyword argument is required for
+  # accessibility reasons and is a keyword argument instead of positional to
+  # match the syntax of `image_tag` which this wraps.
+  def force_https_image_tag(url, alt:)
+    image_tag(force_https(url), alt: alt)
+  end
+
+  # force scheme of a url to https if it is currently http. if it is not http
+  # just return the input with no changes
+  def force_https(url)
+    uri = URI(url)
+    return url unless uri.scheme == 'http'
+    uri.scheme = 'https'
+    uri.to_s
+  end
 end
--- a/app/views/record/_basic_info.html.erb
+++ b/app/views/record/_basic_info.html.erb
@@ -6,7 +6,7 @@
 
 <% if @record.eds_cover_thumb_url.present? %>
 <div class="record-image">
-  <%= image_tag(@record.eds_cover_thumb_url, alt: "Cover for #{@record.title}") %>
+  <%= force_https_image_tag(@record.eds_cover_thumb_url, alt: "Cover for #{@record.title}") %>
 </div>
 <% end %>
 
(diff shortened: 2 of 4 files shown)


app/helpers/record_helper.rb
```diff
@@ -128,4 +128,19 @@ def restricted_link?
   def guest_and_restricted_link?
     guest? && restricted_link?
   end
+
+  # force http images to use https. If remote server cannot handle https a
+  # broken image is expected
```
thatandromeda: Is this also expected *by UX* and they're cool with that?

app/helpers/record_helper.rb
```diff
@@ -128,4 +128,19 @@ def restricted_link?
...
+  def force_https_image_tag(url, alt)
+    image_tag(force_https(url), alt: alt)
```
thatandromeda: Is there a more 🌈  way to do this? Like with this, we have to remember to use the special function every time we put an image in....is there a way to do it automatically?

app/views/record/_basic_info.html.erb
```diff
@@ -6,7 +6,7 @@
 
 <% if @record.eds_cover_thumb_url.present? %>
 <div class="record-image">
-  <%= image_tag(@record.eds_cover_thumb_url, alt: "Cover for #{@record.title}") %>
+  <%= force_https_image_tag(@record.eds_cover_thumb_url, "Cover for #{@record.title}") %>
```
thatandromeda: Are we going to be happier if this acts like `(url, alt: alt)`, so that it has the same syntax as the `image_tag` it replaces?

app/helpers/record_helper.rb
JPrevost: lol. I thought so but I will confirm.

app/helpers/record_helper.rb
JPrevost: Since we use two different methods of calling the API, it seemed best to make it a display concern (for now). If we transition everything to the gem, we may get it for free eventually with https://github.com/ebsco/edsapi-ruby/issues/82

app/views/record/_basic_info.html.erb
JPrevost: You know, I had the same thought and ignored it. I'll reconsider as both of us wondering that is definitely probably a signal we should change it.

app/helpers/record_helper.rb
thatandromeda: OK, cool.

thatandromeda:
(Go ahead and ignore the coveralls warning.)

PR #283, opened 2017-10-26
Prevent 'request scan' button for engineering standards

thatandromeda:
## Status
**READY**

#### What does this PR do?
Prevents display of the "request scan" button for a set of engineering standards that we don't scan.

Compare `/record/cat00916a/mit.001480933` on the PR build and the staging server.

- https://mitlibraries.atlassian.net/browse/DI-569

#### Todo:
- [x] Tests
- [ ] Documentation
- [x] Stakeholder approval

NO

NO


app/models/concerns/standards.rb
```diff
@@ -0,0 +1,37 @@
+module Standards
+  # Just a list of all the Barton item numbers for engineering standards that
+  # we don't scan. They're here so as to avoid cluttering up button_maker.rb.
+  BARTON = ["001480933", "001480931", "001481063", "000604282", "001481545",
```
JPrevost: 🙈

thatandromeda: 🍻

thatandromeda:
FWIW I did add a comment on the ticket that this needs a maintenance plan and Darcy has acknowledged.

I don't love the static list, but I *also* don't love injecting another parameter into the JavaScript, thereby coupling our Aleph-based and EBSCO-based views more tightly (and then having to sanitize it, because it's all searchLink garbage rather than something atomic, and then relying on that searchLink syntax not to change). ¯\_(ツ)_/¯

JPrevost:
Yeah. I think I currently prefer injecting a flag of some sort to this as it would require no maintenance plan. I think we can make it atomic in a Helper and then pass in a boolean to the js. I'm not sold on either yet, I'm just thinking out loud.

PR #285, opened 2017-10-30
update dependencies

JPrevost:
## Status
**READY**

#### What does this PR do?

Updates dependencies

NO

YES


JPrevost:
@frrrances This includes the long awaited update you put in to Kaminari... but I think that's all feature flagged away so it isn't user facing. Can you confirm this either works as intended, is not user facing for now so whatever, or both? Thx!

frrrances:
Not user-facing for now so :shipit:! 🚀

PR #287, opened 2017-11-03
Custom instrumentation of search_target method

JPrevost:
## Status
**IN DEVELOPMENT**

#### What does this PR do?

adds Skylight custom instrumentation to search_target method

the default instrumentation is not providing enough detail to
accurately understand what the slow bits are

Look at the staging vs production skylight apps.

NO

NO


JPrevost:
I don't think this is necessary at this time. This proved (to me at least) that the entire slow bit of this is the remote API calls. Closing for now and if we want to dig in later on anything more specifically we can revisit this.

PR #293, opened 2017-11-17
Exclude scanning button based on Subject Headings

JPrevost:
## Status
**READY**

#### What does this PR do?

The initial work on this feature used a list of Aleph IDs as the
exclusion criteria. After business rules were clarified, it became
desirable to use the same logic we use elsewhere in our systems,
which is the LC Subject Headings of the record.

https://mit-bento-staging-pr-293.herokuapp.com/record/cat00916a/mit.000603781
vs
https://mit-bento-staging.herokuapp.com/record/cat00916a/mit.000603781

Additional example records available in DI-592

This implements the business logic defined in:
- https://mitlibraries.atlassian.net/browse/DI-592

This completes the work in:
- https://mitlibraries.atlassian.net/browse/DI-608

#### Todo:
- [x] Tests
- [x] Documentation
- [x] Stakeholder approval

NO

NO


app/assets/javascripts/search.js
```diff
@@ -16,9 +16,9 @@ function RealtimeStatus( id ) {
   });
 }
 
-function RealtimeItem( an , oclc ) {
+function RealtimeItem( an , oclc, scan ) {
   $.ajax({
-    url: "/full_item_status?id=" + an + '&oclc=' + oclc
+    url: "/full_item_status?id=" + an + '&oclc=' + oclc + '&scan=' + scan
```
thatandromeda: Are we worried that our students will simply edit the URLs when they don't see the 'request scan' button but want to? (I mean, I'm 100% confident this will happen, given this student population, with this implementation. I don't know if it'll happen often enough that we need to care.)

I'm also increasingly uncomfortable about how we're using this URL to couple Aleph and EDS data sources more and more tightly. I didn't like having to incorporate the OCLC number to make it work and now there's yet another parameter. Can we get this data from the Aleph side?

app/helpers/record_helper.rb
```diff
@@ -150,4 +150,24 @@ def force_https(url)
     uri.scheme = 'https'
     uri.to_s
   end
+
+  # Determine scan exclusions based on cataloged subject headings
+  def scan?
+    return true if @subjects.blank?
+    return true if ENV['SCAN_EXCLUSIONS'].blank?
+    if unscannable_subject_heading?
```
thatandromeda: Should this just return `scannable_subject_heading?` rather than having a whole if condition around a boolean to reverse its parity?

config/environments/test.rb
```diff
@@ -20,6 +20,9 @@
   ENV['PER_PAGE'] = '10'
   ENV['FLIPFLOP_KEY'] = 'yoyo'
   ENV['ALEPH_HINT_SOURCE'] = 'https://fake.example.com/s/fake/getserial_mini.xml?dl=1'
+
+  ENV['SCAN_EXCLUSIONS']='Materials -- Standards -- United States -- Periodicals;Materials -- Standards -- United States -- Periodicals;Standards, Engineering;Standards, Engineering -- Periodicals'
```
thatandromeda: IME this is not adequate. The spacing and hyphenation of subjects are not consistent. (This is why in `Rainbows` I actually strip out all the subdivisions and use only the first part of the term, at the cost of some accuracy - it would have taken actual time for me to map the options for subdivision syntax and I wasn't going to do that for an Easter egg. But in this case we actually have to do it.)

JPrevost: Did you see that this was a problem with data coming from Aleph only? This is only a feature for Aleph records Aleph uses controlled subject headings so your concern should be less, uh, concerning?

app/assets/javascripts/search.js
JPrevost: I 100% don't think this is worth worrying about. I totally get it, but if someone really wants to request a scan and then be denied by staff they can indeed do so.

app/helpers/record_helper.rb
JPrevost: Probably. I'll take a look.

config/environments/test.rb
JPrevost: Will confirm with Catalog team that we use controlled authorities for Subjects and report back.

app/assets/javascripts/search.js
JPrevost: Will update in-app documentation to explain why we are doing this mess.

JPrevost: I've added a commit to explain the design decision in the AlephItem class.

config/environments/test.rb
JPrevost: I have confirmation from our catalog team that Subject headings are under authority control and thus should remain as accurate as this expects them to be.

JPrevost:
@thatandromeda I have added commits to address to concerns and confirmed that subject headings are under authority control in our Aleph. Please let me know if you continue to have concerns that prevent merge. Thanks!

thatandromeda:
The coveralls issue is not meaningful.  I am OK with ignoring the codeclimate ones as well, although I don't actually understand why `item_status` and `full_item_status` are different - should they be?

JPrevost:
@thatandromeda the two identical controller methods are rendering two distinct HTML templates. This is definitely not ideal but since it's not new to this PR I'm going to leave it. The alternative is a single method that conditionally loads the correct template and I'm not sure that's any better as it sort of deviates from the rails standard of rendering the template that matches the name of the method. It _can_ be done but I'm not sure it's worth it. Thoughts very much welcome (but I'm merging so any discussion would lead to a future tickets to address).

thatandromeda:
I think adding a comment explaining why there are identical-looking methods suffices. Thanks for the clarification.

JPrevost:
@thatandromeda I added a note to those similar methods and then squashed to merge... but then realized I was not up to date with master. I've rebased but that required updating a test you just added. I think it's a simple squash and merge from here but thought I'd give you another chance to 👍 👎

thatandromeda:
👍

PR #294, opened 2017-11-21
Avoid displaying bogus checked-out item when we exceed Aleph limit

thatandromeda:
See ticket for DI-610.

## Status
**READY**

#### What does this PR do?
Addresses problem where some records were displaying blank (but checked-out) items.

It turns out they weren't items at all, but bogus data returned from our attempt to parse something that was not an item record. The Aleph API only returns 990 items; if there are more than that, then the `<items>` element has a child `<partial>` which warns you that the list has been truncated. Parsing that, while assuming it is an `<item>`, turns out to not really work out for you.

Look at the link in the ticket, on the PR build, and note that it doesn't suck.

- https://mitlibraries.atlassian.net/browse/DI-610

#### Todo:
- [ ] Tests
- [ ] Documentation
- [ ] Stakeholder approval

NO

NO

### DIFF
--- a/app/models/aleph_item.rb
+++ b/app/models/aleph_item.rb
@@ -15,7 +15,12 @@ class AlephItem
   # ex: https://walter.mit.edu/rest-dlf/record/MIT01000293592/items?view=full&key=SECRETKEY
   def items(id, oclc)
     items = []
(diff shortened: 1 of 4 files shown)


thatandromeda:
OK - regression tests added, commits squashed, will merge when checks pass.

PR #295, opened 2017-11-30
Adds an in app toggle link to preview full records

JPrevost:
## Status
**READY**

#### What does this PR do?

* Adds a feature flagged link to enable another feature. Oh my.
* Provides an in app link to enable the full record displays.
* The intent is that once we are ready and enable this feature, users
will be able to self opt-in / opt-out of the full record displays for
some period of time.

Go to the PR build and click madly on the toggle link (and ideally also confirm the results pages change target URLs as appropriate).

- https://mitlibraries.atlassian.net/browse/DI-609

#### Todo:
- [x] Tests
- [ ] Documentation
- [ ] Stakeholder approval

NO

NO

### DIFF
--- a/app/controllers/feature_controller.rb
+++ b/app/controllers/feature_controller.rb
@@ -10,7 +10,7 @@ class FeatureController < ApplicationController
 
   def full_record_toggle
     session[:local_full_record] = session[:local_full_record].!
-    flash[:info] = "In app full record displays are now #{full_record_message}."
+    flash[:info] = "The beta item detail view is now #{full_record_message}."
     redirect_back(fallback_location: root_path)
   end
 
@@ -18,9 +18,9 @@ class FeatureController < ApplicationController
 
   def full_record_message
     if Flipflop.local_full_record?
-      'enabled'
+      'on'
     else
-      'disabled'
+      'off'
     end
   end
 
--- a/app/controllers/feature_controller.rb
+++ b/app/controllers/feature_controller.rb
@@ -1,5 +1,5 @@
 class FeatureController < ApplicationController
-  before_action :validate_feature!
+  before_action :validate_feature!, only: :toggle
 
   # Toggles boolean for session based feature flags
   def toggle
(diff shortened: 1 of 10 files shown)


JPrevost:
@frrrances can you look at the link itself (wording, and styling)
@thatandromeda that rest is all you!

app/controllers/feature_controller.rb
```diff
@@ -8,8 +8,22 @@ def toggle
     redirect_to root_path
   end
 
+  def full_record_toggle
+    session[:local_full_record] = session[:local_full_record].!
+    flash[:info] = "In app full record displays are now #{full_record_message}."
```
thatandromeda: I wonder whether this is the most human-friendly language ("on" vs "off" might be more expected). But maybe it doesn't matter for a thing we don't plan to deploy for super long.

config/environments/test.rb
```diff
@@ -20,6 +20,7 @@
   ENV['PER_PAGE'] = '10'
   ENV['FLIPFLOP_KEY'] = 'yoyo'
   ENV['ALEPH_HINT_SOURCE'] = 'https://fake.example.com/s/fake/getserial_mini.xml?dl=1'
+  ENV['FULL_RECORD_TOGGLE_BUTTON'] = 'true'
```
thatandromeda: Squee!

test/features/toggle_fullrecord_test.rb
```diff
@@ -0,0 +1,24 @@
...
+feature 'ToggleFullrecord' do
+  before do
+    Capybara.current_session.driver.browser.clear_cookies
+  end
+
+  after do
+    Capybara.current_session.driver.browser.clear_cookies
+  end
+
+  test 'can toggle full record from app' do
```
thatandromeda: If I were being super picky I'd test whether the session variable got set. If I were thinking about the fact that the expected deployment time of this is ~2 months before we blow it away, I wouldn't care. Up to you :)

thatandromeda:
LGTM! :shipit: unless 🙈 per Frances.

test/features/toggle_fullrecord_test.rb
JPrevost: It's in the back of my mind to move from session to cookies to persist features across sessions which FlipFlop can support so I'm going to leave this as checking the effects of the change rather than the change itself (for better or worse).

thatandromeda: OK then 🌈

PR #296, opened 2017-12-05
Adds caching for eds-api and aleph responses

JPrevost:
## Status
**READY**

#### What does this PR do?

- adds caching for eds-api and aleph responses
- this uses ActionCaching
http://guides.rubyonrails.org/caching_with_rails.html#action-caching
While ActionCaching has been removed from Rails core, it is still a
valid and viable solution.

- the eds-api data changes infrequently so caching is likely to speed
up reloads and repeat searches
- the aleph data we really need to be "live" so we cache it with a
shorter expiration. However, most data doesn't change so live-live
is overkill and this will speed up refreshes.

View a full record in the PR build from an on-campus IP. It should take a bit of time to load.
Reload the page and it should be super fast. Yay caching.

From a non-campus IP, reload the same record and it should be slower again the first time but fast again on the second load as we cache results dependent on guest status.

- https://mitlibraries.atlassian.net/browse/DI-500

#### Todo:
- [x] Tests
- [x] Documentation
- [ ] Stakeholder approval

NO

YES

### DIFF
--- a/app/controllers/aleph_controller.rb
+++ b/app/controllers/aleph_controller.rb
@@ -1,4 +1,7 @@
 class AlephController < ApplicationController
+  # We are using ActionCaching here because we also use it in EDS full records.
+  # This would work fine with low level caching, but ActiveCache is so clean
+  # and once we are using it someplace else we may as well use it here.
   caches_action :full_item_status, expires_in: 15.minutes,
                                    cache_path: :cache_path
   caches_action :item_status, expires_in: 15.minutes, cache_path: :cache_path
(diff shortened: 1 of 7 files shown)


app/controllers/aleph_controller.rb
```diff
@@ -1,4 +1,8 @@
 class AlephController < ApplicationController
+  caches_action :full_item_status, expires_in: 15.minutes,
```
thatandromeda: This is awesomely easy to read.

thatandromeda:
Can this have a comment somewhere logical explaining the decision behind using ActionCaching? Once you add the comment go ahead and merge, LGTM. Yay speed!

As a side note, I have a strong preference for comments that are full sentences, starting in a capital letter and ending in a period (I find enforcing that rule on myself makes me write much clearer and more explanatory comments). Not going to block on that, though.

JPrevost:
@thatandromeda added some notes on our decision to use ActiveCache to where we use it as you suggested and will now merge.

PR #302, opened 2017-12-13
Scan button bug fixes

JPrevost:
## Status
**READY**

#### What does this PR do?
- removes double encoding of passed title and changes to
`URI.encode_www_form_component`
- adds `rft.date` to populate year field
- adds `rft.volume` to populate volume field in the future (changes to
SFX script will be necessary to get this to pass through to Illiad)

On the PR build, activate full record views.

Search for a local record and click "request scan".

The year should be populated and the title should not be encoded.

Confirm on the staging server the same record leads to no year being populated and the title being encoded.

- https://mitlibraries.atlassian.net/browse/DI-618
- https://mitlibraries.atlassian.net/browse/DI-619
- prep for https://mitlibraries.atlassian.net/browse/DI-623

#### Todo:
- [x] Tests
- [x] Documentation
- [x] Stakeholder approval

NO

NO

### DIFF
--- a/app/concerns/sfx_handler.rb
+++ b/app/concerns/sfx_handler.rb
@@ -1,5 +1,4 @@
 class SFXHandler
-
   # The SFX server will do the best it can with whatever it has, so we will
   # initialize the SFXHandler with whatever data we happen to have. We're not
   # guaranteed to have all (or indeed any) of these objects; Aleph and EDS
@@ -11,13 +10,17 @@ class SFXHandler
                  collection: nil,
                  doc_number: nil,
                  library: nil,
-                 title: nil)
+                 title: nil,
+                 year: nil,
+                 volume: nil)
     @barcode = barcode
     @call_number = call_number
     @collection = collection
     @doc_number = doc_number
     @title = title
     @library = library
+    @year = year
+    @volume = volume
   end
 
   # Use this when you want to request a scan of an object.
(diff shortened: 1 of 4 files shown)


JPrevost:
@thatandromeda this requires a new review as it is a reimplementation of the previous PR more than a simple rebase due to the changes that landed in master prior to this work.

app/concerns/sfx_handler.rb
```diff
@@ -54,15 +52,17 @@ def url_constructor(scan: false)
...
+    encoded_call_no = URI.encode_www_form_component(@call_number)
 
     url_parts = [
       sfx_host.to_s,
       "?sid=ALEPH:#{analytics}",
       "&amp;call_number=#{encoded_call_no}",
       "&amp;barcode=#{@barcode}",
-      "&amp;title=#{encoded_title}",
-      "&amp;location=#{encoded_location}"
+      "&amp;title=#{URI.encode_www_form_component(@title)}",
```
thatandromeda: I don't particularly care if we do the encoding here during the interpolation or earlier, but it seems like call_no and title should be handled the same way.

JPrevost: I didn't move the call number in because of the documentation comment about it whereas the title had no comments so it felt okay to just move it inline.

JPrevost:
@thatandromeda I've added a commit that normalizes how we construct the values necessary for the url construction.

thatandromeda:
Oh, that's much better. 👍  LGTM.

PR #315, opened 2018-01-31
Bug fix in bento view pub types

JPrevost:
## Status
**READY**

#### What does this PR do?

Bug fix for handling of multiple pub types in bento view

search for `popcorn` on staging and this PR. On staging you'll get an error in the books bucket. In this PR you should get results. Note that one of the results has no Publication Type... but it isn't broken now at least. 🙈 

- https://mitlibraries.atlassian.net/browse/DI-662

#### Todo:
- [x] Tests
- [x] Documentation
- [ ] Stakeholder approval

NO

NO

### DIFF
--- a/app/models/normalize_eds_common.rb
+++ b/app/models/normalize_eds_common.rb
@@ -89,9 +89,11 @@ class NormalizeEdsCommon
   # full record view, here we're working with our own Frankenmodel and we have
   # to rebuild the affordance.
   def type
-    typepub = @record['Items'].select do |hash|
-      hash['Name'] == 'TypePub'
-    end.first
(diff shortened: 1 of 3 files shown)


thatandromeda:
It's weird to me that this results in cases where no item type displays, because the last part of the conditional is just the logic we had before. Do we also need a `&& typepub['Data'].present?` before we return `typepub['Data']`?

Also um 🙈  on the original bug...I should have caught that. Sorry.

JPrevost:
It's not a bug with our logic. It's library metadata at it's finest.

It's like this in prod as well before this code was introduced.

https://lib.mit.edu/search/bento?q=popcorn

PR #317, opened 2018-01-31
660 buttons are the worst

JPrevost:
## Status
**READY**

#### What does this PR do?

This adds the z30status of the items that had a button we didn't want to the badz30 status. I'm not entirely sure this is a good idea. It fixed the problem for the sample records but I would like a solid opinion on whether this was a good idea from the great ButtonMaker perspective.

Putting so much questionable business logic into these buttons was probably not the best idea we ever had.

Check the records noted in the ticket in staging and the PR build

- https://mitlibraries.atlassian.net/browse/DI-660

#### Todo:
- [ ] Tests
- [ ] Documentation
- [ ] Stakeholder approval

NO

NO

### DIFF
--- a/app/models/button_maker.rb
+++ b/app/models/button_maker.rb
@@ -95,7 +95,8 @@ class ButtonMaker
       'Music CD/DVD',
       'Pass',
       'See Note Above',
-      'Two Week Loan'
+      'Two Week Loan',
+      'One Week Loan'
     ]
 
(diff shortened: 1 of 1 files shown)


thatandromeda:
I share your concerns and I left a comment in the ticket to confirm the business logic side. So far the code looks 🌈  but I want to hold off on approval until I have that stakeholder confirmation. You can ignore the 0.003% coverage decrease ;)

JPrevost:
grumble grumble

PR #320, opened 2018-02-02
Refactor buttons

JPrevost:
## Status
**READY**

#### What does this PR do?

Moves common button methods to a module and unique elements to independent classes in the hopes of making a more easily reasoned implementation of our complex business rules.

Hold and Recall buttons remain so tightly coupled that they share a class even in this refactor.

I was careful to keep all existing tests and just changed method calls or stubs as appropriate to match the new classes. However, see the note below about how some of the tests don't seem to work as expected after the refactor and need additional work.

NOT DONE YET: 

- [x] a few tests pass but are labeled with `todo` flags because the instance variable they are using to change intended test outcomes was removed but they unfortunately still pass. I will be adding additional tests to cover their intended purpose before removing those in a separate commit but I wanted to provide an opportunity for discussion as soon as possible.
- [x] review all method documentation to ensure it still makes sense with this refactor
- [x] update docs on `ButtonMaker` modules to explain how the the Classes that include `ButtonMaker` are intended to work (see pr comment from Andromeda for better explanation)
- [x] In hold/recall tests ensure whenever we assert there is a hold button we refute there is a recall button and vice versa (i.e. we should always only have one of these two at most... zero is fine)
- [x] document `all_buttons` to explain this is where the magic starts and the order matters as to how they display in the UI
- [x] `available_here_now?` is better than `available_locally?`
- [x] Make it clear in docs why `ButtonHoldRecall` can only spit out one button and why they share one class unlike the other buttons.

There should be no user facing changes. This is all read the code stuff. Before merging I will have stakeholders test the PR build fully.

- https://mitlibraries.atlassian.net/browse/DI-666

#### Todo:
- [ ] Tests
- [ ] Documentation
- [ ] Stakeholder approval

NO

NO

### DIFF
--- a/app/models/aleph_item.rb
+++ b/app/models/aleph_item.rb
@@ -39,7 +39,7 @@ class AlephItem
   # Note that it swaps b <=>a for description but a<=>b for library. This lets
   # us display dates/volumes in DESCENDING order (on the theory that people are
   # most likely looking for more recent holdings), but libraries in ASCENDING
-  # order (since this is how humans expect alphabetization to work). 
+  # order (since this is how humans expect alphabetization to work).
   def custom_sort(items)
     items.sort do |a, b|
       [b[:description][/\d+/].to_i, a[:library]] <=>
@@ -54,7 +54,14 @@ class AlephItem
       available?: available?(item),
       label: label(item),
       description: description(item),
-      buttons: ButtonMaker.new(item, oclc, scan).all_buttons }
+      buttons: all_buttons(item, oclc, scan) }
+  end
+
+  def all_buttons(item, oclc, scan)
+    [ButtonContact.new(item, oclc, scan).html_button,
+     ButtonHoldRecall.new(item, oclc, scan).html_button,
+     ButtonIll.new(item, oclc, scan).html_button,
+     ButtonScan.new(item, oclc, scan).html_button]
   end
 
   # ~~~~~~~~~~~~~~~~~~~~~~~~ Properties of Aleph items ~~~~~~~~~~~~~~~~~~~~~~~~
--- a/app/models/aleph_item.rb
+++ b/app/models/aleph_item.rb
@@ -57,6 +57,10 @@ class AlephItem
       buttons: all_buttons(item, oclc, scan) }
   end
 
+  # Defines the buttons and button order that may display. Each button has
+  # logic that determines whether it will output HTML or not, so all we need to
+  # do here is call the #html_button method of each button we are interested in
+  # and the button will take care of the rest.
   def all_buttons(item, oclc, scan)
     [ButtonContact.new(item, oclc, scan).html_button,
      ButtonHoldRecall.new(item, oclc, scan).html_button,
--- a/app/models/button_contact.rb
+++ b/app/models/button_contact.rb
@@ -1,3 +1,5 @@
+# Creates a Button with an actionable URL for items that require contacting
+# a library for more information.
 class ButtonContact
   include ButtonMaker
 
--- a/app/models/button_contact.rb
+++ b/app/models/button_contact.rb
@@ -1,0 +1,13 @@
+class ButtonContact
+  include ButtonMaker
+
+  def html_button
+    return unless eligible?
+    "<a class='btn button-secondary button-small' " \
+      "href='https://libraries.mit.edu/archives/'>Contact Us</a>"
+  end
+
+  def eligible?
+    @library == 'Institute Archives'
+  end
+end
--- a/app/models/button_maker.rb
+++ b/app/models/button_maker.rb
@@ -44,17 +44,12 @@ module ButtonMaker
     @volume = @item.xpath('z30/z30-description').text
   end
 
-  # The following functions construct URLs needed for item availability actions.
-  # WE NEED THESE EVEN IF WE REFACTOR TO RELY MORE ON EBSCO RESULTS - they
-  # require Aleph data which is not known to EBSCO.
-  # These URLs will present authentication challenges for non-logged-in users,
-  # which will redirect appropriately upon success.
-
   # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ Misc utilities ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-  # We won't generally ILL items that fit these criteria - we want them to be
-  # placed on hold if possible.
-  def available_locally?
+  # This is used for both Hold/Recall and ILL purposes.
+  # We won't generally ILL or Recall items that fit these criteria as we want
+  # them to be placed on hold if possible.
+  def available_here_now?
     # You can request things that are in the library and have reasonable
     # loan policies.
     [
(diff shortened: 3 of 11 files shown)


app/models/aleph_item.rb
```diff
@@ -54,7 +54,14 @@ def process_item(item, oclc, scan)
       available?: available?(item),
       label: label(item),
       description: description(item),
-      buttons: ButtonMaker.new(item, oclc, scan).all_buttons }
+      buttons: all_buttons(item, oclc, scan) }
+  end
+
+  def all_buttons(item, oclc, scan)
+    [ButtonContact.new(item, oclc, scan).html_button,
+     ButtonHoldRecall.new(item, oclc, scan).html_button,
```
thatandromeda: There is ~absolutely no~ circumstance where we could end up with both a hold button and a recall button, right? Like, this doesn't need to be something like
```
ButtonHoldRecall.new(item, oclc, scan, mode='hold').html_button,
ButtonHoldRecall.new(item, oclc, scan, mode='recall').html_button,
```
?

(I will duly glare at myself for having used modes here.)

app/models/aleph_item.rb
```diff
@@ -54,7 +54,14 @@ def process_item(item, oclc, scan)
       available?: available?(item),
       label: label(item),
       description: description(item),
-      buttons: ButtonMaker.new(item, oclc, scan).all_buttons }
+      buttons: all_buttons(item, oclc, scan) }
+  end
+
+  def all_buttons(item, oclc, scan)
```
thatandromeda: There should be some comment reminding future-us that if we add any types of button, we MUST LIST THEM HERE or else it's not-🌈 .

app/models/aleph_item.rb
JPrevost: Ha. I was thinking of adding in tests for this explicitly. In the tests whenever we check for one check the other is the opposite. I'll add that to the todos (I'm confident this is a non-issue but the tests would be super easy to add so why not)

app/models/button_contact.rb
```diff
@@ -0,0 +1,13 @@
+class ButtonContact
+  include ButtonMaker
+
+  def html_button
+    return unless eligible?
+    "<a class='btn button-secondary button-small' " \
+      "href='https://libraries.mit.edu/archives/'>Contact Us</a>"
+  end
+
+  def eligible?
```
thatandromeda: ahhhhhh yeah I like so much how this refactor lets us have a simple `eligible?` interface.

app/models/button_maker.rb
```diff
@@ -51,242 +44,17 @@ def set_item_properties
...
-      # The Annex is special.
-      %w(13 14 15 56 57).include?(@z30status_code) &&
-        @library == 'Library Storage Annex'
-    ].any?
-  end
-
   # We won't generally ILL items that fit these criteria - we want them to be
   # placed on hold if possible.
-  def should_you_get_it_here?
+  def available_locally?
```
thatandromeda: I do think that a somewhat different name is better, per discussion on the other PR - the key thing is that it's available both here _and now_ (as availability is very time-dependent). `available_here_now?` ?

app/models/button_maker.rb
```diff
@@ -17,19 +17,12 @@
 # pulling it all into a special-purpose, encapsulated place.
 # You probably don't want to, but when all else fails familiarize yourself with:
 # https://mitlibraries.atlassian.net/wiki/spaces/DI/pages/58654721/ButtonMaker+documentation
-class ButtonMaker
+module ButtonMaker
```
thatandromeda: Now that we've factored out all kinds of stuff (🎉 ), there should be some kind of comment here as to the types of interfaces child classes are expected to expose - in particular, that they must all define `html_button`, which returns either `nil` or HTML suitable for enabling this function, and they must all determine whether to return the button or `nil` (using a function `eligible?` unless there's some really good reason for it to be more complicated than that). And if there's anything interesting about why functionality is here rather than in child classes, that should also be addressed (although on first glance maybe the separation of functions is pretty logical and doesn't require explanation).

app/models/button_maker.rb
JPrevost: 👍

app/models/button_contact.rb
JPrevost: Except on the HoldRecall :/ I tried to make that the same interface but it was too big of a refactor to keep the tests reasonable static for these changes.

app/models/aleph_item.rb
thatandromeda: 👍

app/models/button_contact.rb
thatandromeda: Yeah :/ I hear you and I think it's cool once documented.

thatandromeda:
Thus far this is super great - I think it'll make our lives much easier and I hadn't seen how to do this so I'm glad you have. Can you point to one or more specific misbehaving tests so I can ponder them?

app/models/aleph_item.rb
JPrevost: Will do... plus the extra tests I noted.

test/models/button_hold_recall_test.rb
```diff
@@ -0,0 +1,164 @@
...
+    maker = ButtonHoldRecall.new(@item, @oclc, @scan)
+
+    # Check our assumption that the default ButtonMaker can be held/recalled.
+    assert maker.eligible_hold?
+
+    # Now test z30 status codes that should yield ineligibility.
+    maker.instance_variable_set(:@z30status_code, '04')
+    # We need to reset @hold_recallable as it was set on object initialization, but
+    # the change to the z30 status will alter its expected value.
+    # todo: hold_recallable is no longer an instance variable. refactor tests.
```
JPrevost: @thatandromeda this is one example of the tests that really should probably fail. I haven't thought much about it as I wanted to get feedback on the general approach of this refactor before coming back and cleaning up the few bits that are just weird.

JPrevost: In other words, I am fairly confident they are no longer actually testing anything useful. I plan to dig in on what they were testing and write new tests to cover those situations.

app/models/aleph_item.rb
thatandromeda: Those extra tests sound great.

test/models/button_hold_recall_test.rb
thatandromeda: 👍

thatandromeda: This looks right? If the code is `04`, `hold_recallable?` should be `false`, so  `eligible_hold?` should be false.

But I feel fine about rethinking what they're actually testing and updating accordingly. At any rate, I don't see anything obviously weird going on here.

JPrevost:
@thatandromeda I believe I have completed the additional documentation and testing we discussed for this PR. Let me know what you think on the code side and I'll get a confirmation from stakeholders it seems to still work for real as well.

thatandromeda:
Looks 🌈 to me. Thanks for this - our lives will be easier if we need to interact with this in future.

PR #325, opened 2018-03-01
Make type listed in document type section match that's indicated up top

thatandromeda:
Testing this feature revealed that the TypePub strategy sometimes
returns nil, so we have a new helper function that uses TypePub info
when present, but falls back to the EDS-supplied publication type
when needed.

## Status
**READY**

#### What does this PR do?
(per commit message)

Compare https://mit-bento-staging.herokuapp.com/record/cat00916a/mit.001512948 to PR build version of same.

- https://mitlibraries.atlassian.net/browse/DI-681

#### Todo:
- [x] Tests
- [ ] Documentation
- [x] Stakeholder approval

NO

NO

### DIFF
--- a/app/views/record/_extended_info.html.erb
+++ b/app/views/record/_extended_info.html.erb
@@ -6,9 +6,9 @@
 
 <h3 class="section-title">More information</h3>
 <ul class="list-moreinfo">
-  <% if @record.eds_publication_type.present? %>
+  <% if publication_type(@record).present? %>
     <li>
-      <span class="label">Document type:</span> <%= @record.eds_publication_type %>
+      <span class="label">Document type:</span> <%= publication_type(@record) %>
     </li>
   <% end %>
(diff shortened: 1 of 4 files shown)


app/views/record/_extended_info.html.erb
```diff
@@ -6,9 +6,11 @@
 
 <h3 class="section-title">More information</h3>
 <ul class="list-moreinfo">
-  <% if @record.eds_publication_type.present? %>
+  <%# Use get_item_data({name:'TypePub'}), not eds_publication_info, to match
```
JPrevost: I'm not sure this comment helps as the info it talks about is actually in the helper... I recommend just not documenting this as the helper is clear enough as to what it does.

thatandromeda:
You're quite right - that comment went with an older version and should have been removed.

PR #345, opened 2018-03-20
record cache key now includes pride enabled status

JPrevost:
## Status
**READY**

#### What does this PR do?

Includes the current users pride feature status as part of the full record view cache key.

Not including this in the cache key resulted in one user enabling the
feature and any subsequent users seeing the rainbows if they loaded the
exact same full record view because the cache was loaded before determining if the user had the feature enabled (because it was not part of the key).

Some things are hard.

In the PR build, enable pride and look at a record that should be affected (see rainbows).

Disable pride and look at the record. It should now reflect a change (no rainbows).

In a private browser session, load the record. You should see no rainbows.

Compare to the same process on current staging env.

er... nope.

NO

NO


test/controllers/record_controller_test.rb
```diff
@@ -67,7 +67,7 @@ class RecordControllerTest < ActionDispatch::IntegrationTest
     VCR.use_cassette('record: bananas', allow_playback_repeats: true) do
       get record_url('cat00916a', 'mit.001492509')
       assert_equal(@controller.send(:cache_path),
-                   'http://www.example.com/record/cat00916a/mit.001492509?guest=true&source=cat00916a')
+                   'http://www.example.com/record/cat00916a/mit.001492509?guest=true&pride=false&source=cat00916a')
```
thatandromeda: Does this mean there could potentially be a conflict between the Flipflop thing and the URL params? What happens if I say `pride=true` in the URL but `Flipflop.enabled?(:pride)` is false (or vice versa)?

JPrevost: `pride=true` is never in the url of an actual request to a record. It's just set once to enable in the user session. The cache key, while it definitely _looks_ like a URL is actually just made that way for convenience. It's really just a string that look like a URL. Even if it _is_ put into the URL manually, the value does _not_ come from the URL params when generating a cache key. It is set directly from the `Flipflop.enabled?(:pride)` boolean response.

However, the original `rainbows=heckyeah` easter egg is in conflict with the pride feature. I think that is best solved by not sharing that direct rainbows easter egg with anyone rather than trying to handle it because omfg I don't care :)

thatandromeda: Ahhhh ok awesome. :shipit:

PR #354, opened 2018-03-29
csv was not being treated as utf8

JPrevost:
## Status
**READY**

#### What does this PR do?

Google Sheets exports as UTF8. We were not treating the file as UTF8.

I’m not sure why this worked before and didn’t anymore.

The failing record was for Encyclopædia Britannica online which loaded fine with this change on the staging server.

NO

NO

### DIFF
--- a/app/models/custom_hint.rb
+++ b/app/models/custom_hint.rb
@@ -45,7 +45,7 @@ class CustomHint
 
   # Load csv hint source.
   def csv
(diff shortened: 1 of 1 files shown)


JPrevost:
Yeah, I'm not sure what changed to require this. I did confirm Google Sheets exports as UTF-8 so the forced encoding is indeed correct so I kinda stopped caring at that point.

PR #403, opened 2018-06-11
Enable ENV toggle-able EDS Debug mode

JPrevost:
## Status
**IN DEVELOPMENT**

#### What does this PR do?

- Setting `EDS_DEBUG` will enable the EDS debug mode
- When EDS API is having session issues, the gem fails with a NoMethodError. This detects that and recovers by supplying users an alternative interface link and logs the problem so we can be sure we understand how often this is happening.

NO

NO

### DIFF
--- a/app/views/errors/eds_session_error.erb
+++ b/app/views/errors/eds_session_error.erb
@@ -1,0 +1,4 @@
+<h2 class="title title-page">An error occurred accessing the detail page you requested.</h2>
+
+<p>You may be able to access similar information this <%= link_to('alternative interface', @eds_error_link) %>.
+  If not, please <%= link_to('let us know', feedback_path) %>. Sorry for the inconvenience this has caused.</p>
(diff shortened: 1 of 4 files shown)


app/views/errors/eds_session_error.erb
```diff
@@ -0,0 +1,4 @@
+<h2 class="title title-page">An error occurred accessing the detail page you requested.</h2>
+
+<p>You may be able to access similar information this <%= link_to('alternative interface', @eds_error_link) %>.
```
thatandromeda: This is missing a word ("..._at_ this alternative interface"?)

app/views/errors/eds_session_error.erb
```diff
@@ -0,0 +1,4 @@
...
+  If not, please <%= link_to('let us know', feedback_path) %>. Sorry for the inconvenience this has caused.</p>
```
thatandromeda: ++

app/views/errors/eds_session_error.erb
JPrevost: Thx. Will address with https://mitlibraries.atlassian.net/browse/DI-720

PR #409, opened 2018-06-21
Handle different failed EDS record responses

JPrevost:
## Status
**READY**

#### What does this PR do?

We were handling all EDS record failures the same, but we have seen two distinct cases in production so this handles those more clearly. We will also continue to handle the generic case to ensure we are seeing in our logs when all of these types happen and respond as appropriate.

- https://mitlibraries.atlassian.net/browse/DI-721

NO

NO

### DIFF
--- a/app/controllers/record_controller.rb
+++ b/app/controllers/record_controller.rb
@@ -1,11 +1,10 @@
 class RecordController < ApplicationController
   before_action :restricted!, only: [:direct_link]
   before_action :valid_url!
-  rescue_from EBSCO::EDS::BadRequest, with: proc {
-    raise RecordController::NoSuchRecordError, 'Record not found'
-  }
 
   class NoSuchRecordError < StandardError; end
+  class DbLimitReached < StandardError; end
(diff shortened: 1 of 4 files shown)


app/controllers/record_controller.rb
```diff
@@ -93,6 +94,20 @@ def with_session_error_handling
     @eds_error_link = rebuild_eds_full_record_link
   end
 
+  # detect and log known and unknown eds failures separately so we can
+  # better understand what is happening
+  def with_error_handling
+    yield
+  rescue EBSCO::EDS::BadRequest => e
+    if e.message.include?('Simultaneous User Limit Reached')
```
thatandromeda: I'm OK with this for now, but it will rapidly turn into twisty spaghetti if we discover more error conditions we want to rescue specifically. Long-term this may be happier as a hash mapping error messages to exceptions or something.

JPrevost: Good idea. If it gets bigger than these cases I think it should get that suggested refactor. 👍

PR #435, opened 2018-08-15
Changes custom EDS API calls to use persistent connections

JPrevost:
## Status
**READY**

#### What does this PR do?

This uses a persistent connection within a single rails request.

Since the EDS API calls require 4 calls to make a single search this should reduce overheard for doing that.

#### Useful background info

https://github.com/httprb/http/wiki/Persistent-Connections-%28keep-alive%29

🤷‍♀️ 

NO

NO

### DIFF
--- a/app/models/search_eds.rb
+++ b/app/models/search_eds.rb
@@ -18,6 +18,7 @@ class SearchEds
 
   def initialize
     @attempts = 0
+    @eds_http = HTTP.persistent(EDS_URL)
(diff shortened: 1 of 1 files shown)


JPrevost:
Thanks for the review (and idea to look into doing this) @gravesm . I've asked the disco UI team to do some testing. If I don't hear back anything negative from them by tomorrow morning I'll merge and push to prod and see if it helps.

PR #437, opened 2018-08-16
Enable puma workers and preload app

JPrevost:
## Status
**READY**

#### What does this PR do?

Enables puma workers and preloads app.

https://devcenter.heroku.com/articles/deploying-rails-applications-with-the-puma-web-server

NO

NO

### DIFF
--- a/config/puma.rb
+++ b/config/puma.rb
@@ -21,7 +21,7 @@ environment ENV.fetch("RAILS_ENV") { "development" }
 # Workers do not work on JRuby or Windows (both of which do not support
 # processes).
 #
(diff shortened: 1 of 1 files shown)


JPrevost:
@depfu rebase

gravesm:
I'm not convinced preloading the app is worth it. My own admittedly limited initial testing shows the worker processes in the preload scenario pretty quickly end up needing their own memory space. I think you'd probably be better off just going with the lighter weight daemon process. I'm going to go ahead and approve, but I'd strongly encourage you to collect some solid memory usage statistics under both scenarios before deciding which one to go with.

PR #472, opened 2018-10-18
Moves to mitlibraries-theme gem

JPrevost:
## Status
**HOLD**

#### What does this PR do?

Moves to the mitlibraries-theme gem

NO

YES


JPrevost:
That test failure is real. I'll fixup.

PR #568, opened 2019-07-31
Removes all dropbox requirements for hints

JPrevost:
## Status
**READY**

#### What does this PR do?

Removes all dropbox requirements for hint loading

NO

NO

### DIFF
--- a/test/models/custom_hint_test.rb
+++ b/test/models/custom_hint_test.rb
@@ -4,7 +4,7 @@ require 'test_helper'
 class CustomHintTest < ActiveSupport::TestCase
 
   setup do
-    @url = 'https://www.dropbox.com/s/f6p4qtj3chaqlp2/test_popular.csv?dl=0'
+    @url = 'https://example.com/whatever.csv'
   end
 
   test 'fails without URL argument' do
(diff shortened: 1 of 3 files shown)


test/models/custom_hint_test.rb
```diff
@@ -106,7 +67,7 @@ class CustomHintTest < ActiveSupport::TestCase
 
   test 'adds fingerprint if not present' do
     VCR.use_cassette('custom hint blank fingerprints', allow_playback_repeats: true) do
-      url = 'https://www.dropbox.com/s/3dgxge8b14ht0c3/custom%20hint%20for%20test%20suite.csv?dl=0'
+      url = 'https://www.dropbox.com/s/3dgxge8b14ht0c3/custom%20hint%20for%20test%20suite.csv?dl=1'
```
matt-bernhardt: I'm trying to understand whether this change needs a corresponding change in a cassette that I'm not seeing?

JPrevost: The cassettes were looking for dl=1 but we removed that code that automatically changed that from dl=0 so I just updated the test instead of the cassette.

matt-bernhardt: Ah, gotcha

PR #600, opened 2019-10-24
Prefer custom links to constructed links and improve contructed links

JPrevost:
#### What does this PR do?

- uses relevant links from EDS custom links section if they exist and fulltext links is not present. This is the second to last option so it is only used when we'd previously construct an SFX link
- adds rft.stitle to our constructed sfx link fallback if present

The logic behind the links is awful and unfortunately we essentially duplicate it because our initial search does direct API calls and our single record view uses an EDS maintained wrapper gem. At one point a project was proposed to normalize that to always using the wrapper gem but that work was denied. It has been long enough and the effects of that decision are painful enough that it may be appropriate to do that work as routine maintenance.

`/record/edsoro/oro.9780195156003.013.0057`

- https://mitlibraries.atlassian.net/browse/DI-801

#### Todo:
- [ ] Tests
- [ ] Documentation
- [x] Stakeholder approval

NO

NO

### DIFF
--- a/app/helpers/record_helper.rb
+++ b/app/helpers/record_helper.rb
@@ -67,6 +67,17 @@ module RecordHelper
     )
   end
 
+  # If there are relevant custom links, use one.
+  def check_custom_link?
+    return false unless @record.fulltext_links.present?
+    @record.fulltext_links.each do |ftl|
+      if relevant_fulltext_link?(ftl)
+        @sfx_link = ftl[:url]
+      end
+    end
+    @sfx_link
+  end
+
   # domains we consider relevant when evaluating links
   def relevant_links
     ['libproxy.mit.edu', 'library.mit.edu', 'sfx.mit.edu', 'owens.mit.edu',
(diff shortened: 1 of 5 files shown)


matt-bernhardt:
(note - this is not a review yet)
I've looked over the changes in the diff, and it seems legible at a superficial level. Understanding that there's no burning pressure to merge this, I want to get more up to speed on this portion of the application before trying to pretend I can meaningfully weigh in.

app/helpers/record_helper.rb
```diff
@@ -190,11 +200,18 @@ def availability_action
...
+
     # When all else fails, make SFX sort it out.
     else
       @sfx_link = SFXHandler.new(
         title: @record.eds_title,
-        doc_number: clean_an,
+        source_title: @record.source_title,
+        year: @record.eds_publication_year,
+        volume: @record.eds_volume,
+        doc_number: clean_an
```
matt-bernhardt: I think I've got a handle on the rest of this PR, but I'm looking at the addition of `year` and `volume` to this, and assuming that I'm missing something. I see them already in the SFXHandler, so are we just starting to pass them from here (while they're already passed from other places that invoke SFXHandler) ?

JPrevost: SFXHandler uses these values if we pass them in... but we weren't paying them in here so it wasn't using them. We should have been passing them in as they are useful bits of information.

PR #632, opened 2020-01-16
Adds ArchivesSpace search panel via the TIMDEX! GraphQL endpoint

matt-bernhardt:
## Status
**IN DEVELOPMENT**

#### What does this PR do?
This PR will probably end up getting deleted, or at least heavily rebased. Mostly, I'm impatient to poke around a bit and get back into this codebase.

- https://mitlibraries.atlassian.net/browse/DI-810 (graphQL syntax)
- https://mitlibraries.atlassian.net/browse/DI-811 (model/tests)
- https://mitlibraries.atlassian.net/browse/DI-813 (views/controller)
- https://mitlibraries.atlassian.net/browse/DI-822 (fields being used)
- https://mitlibraries.atlassian.net/browse/DI-823 (app.json structure)

#### Todo:
- [x] Tests
- [ ] Documentation
- [x] Stakeholder approval

NO

NO

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,8 @@ custom hint metadata.
   If exceeded, 'et al' will be appended after this number.
 - `RECAPTCHA_SITE_KEY`
 - `RECAPTCHA_SECRET_KEY`
+- `TIMDEX_URL`: The GraphQL endpoint for Timdex/ArchiveSpace
+- `ASPACE_SEARCH_URI`: the base search URL for the Timdex/ArchiveSpace application
 
 ## Optional Environment Variables
 
--- a/app/assets/stylesheets/_results.scss
+++ b/app/assets/stylesheets/_results.scss
@@ -70,6 +70,11 @@
      color: $gray;
    }
 
+  .result-physical {
+    font-size: $fs-xsmall;
+    color: $gray;
+  }
+
   .result-pubinfo {
     font-size: $fs-xsmall;
     color: $gray;
@@ -83,6 +88,10 @@
     &:before {
       content: " | ";
       color: $gray-l2;
+    }
+
+    &:first-child:before {
+      content: none;
     }
    }
 
--- a/app/models/normalize_timdex.rb
+++ b/app/models/normalize_timdex.rb
@@ -1,0 +1,54 @@
+# Tranforms results from {SearchTimdex} into normalized {Result}s
+#
+class NormalizeTimdex
+  # Translate Timdex results into local result model
+  def to_result(results, q)
+    norm = {}
+    norm['total'] = results['data']['search']['hits']
+    norm['results'] = []
+    norm['eds_ui_view_more'] = view_more(q)
+    extract_results(results, norm)
+    norm
+  end
+
+  private
+
+  def view_more(q)
+    "https://archivesspace.mit.edu/search?op%5B%5D=&q%5B%5D=#{q}"
+  end
+
+  def extract_results(results, norm)
+    return unless results['data']['search']['records']
+    results['data']['search']['records'].first(5).each do |record|
+      result_title = extract_title(record)
+      result = Result.new(result_title, record['sourceLink'])
+      result.blurb = extract_first_from_list(record['summary'])
+      result.year = extract_field(record['publicationDate'])
+      result.authors = extract_field(extract_list(record['contributors']))
+      result.physical_description = extract_field(record['physicalDescription'])
+      norm['results'] << result
+    end
+  end
+
+  def extract_title(record)
+    title = record['title']
+    if record['identifier'].present?
+      title << ' (' + record['identifier'].to_s.gsub('MIT:archivespace:','') + ')'
+    end
+    return title
+  end
+
+  def extract_list(contributors)
+    contributors&.map do |creator|
+      [creator['value'], "&field[]=creators_text&q[]=" << URI.encode_www_form_component(creator['value'])]
+    end
+  end
+
+  def extract_field(value)
+    ( value.present? ) ? value : false
+  end
+
+  def extract_first_from_list(list)
+    ( list.present? ) ? list[0] : false
+  end
+end
--- a/app/models/search_timdex.rb
+++ b/app/models/search_timdex.rb
@@ -1,0 +1,31 @@
+# Searches TIMDEX! API and formats results into {Result} object
+#
+# == Required Environment Variables:
+# - TIMDEX_URL
+#
+# == See Also
+# - https://timdex.mit.edu/playground
+class SearchTimdex
+  attr_reader :results
+
+  TIMDEX_URL = ENV['TIMDEX_URL'].freeze
+
+  def initialize
+    @timdex_http = HTTP.persistent(TIMDEX_URL)
+      .headers(accept: 'application/json',
+        'Accept-Encoding': 'gzip, deflate, br',
+        'Content-Type': 'application/json',
+        'Origin': 'https://lib.mit.edu'
+      )
+    @results = {}
+  end
+
+  # Run a search
+  # @param term [string] The string we are searching for
+  # @return [Hash] A Hash with search metadata and an Array of {Result}s
+  def search(term)
+    @query = '{"query":"{search(searchterm: \"' + term + '\", source: \"MIT ArchivesSpace\") {hits records {sourceLink title identifier publicationDate physicalDescription summary contributors { value } } } }"}'
+    results = @timdex_http.post(TIMDEX_URL, :body => @query)
+    json_result = JSON.parse(results.to_s)
+  end
+end
--- a/app/views/search/_result.html.erb
+++ b/app/views/search/_result.html.erb
@@ -1,6 +1,8 @@
 <div class="wrap-result">
   <% if Flipflop.enabled?(:local_browse) %>
     <% search_prefix = '/search/bento?q=' %>
+  <% elsif 'timdex' == params[:target] %>
+    <% search_prefix = ENV['ASPACE_SEARCH_URI'] %>
   <% else %>
     <% search_prefix = ENV['EDS_PROFILE_URI'] %>
   <% end %>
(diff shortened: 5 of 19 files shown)


app/models/search_timdex.rb
```diff
@@ -0,0 +1,35 @@
...
+  def initialize
+    @timdex_http = HTTP.persistent(TIMDEX_URL)
+    @results = {}
+  end
+
+  # Run a search
+  # @param term [string] The string we are searching for
+  # @return [Hash] A Hash with search metadata and an Array of {Result}s
+  def search(term)
+    @query = '{"query":"{\n  search(searchterm: \"' + term + '\", source: \"MIT ArchivesSpace\") {\nhits\n    records {\n      sourceLink\n      title\n    }\n  }\n}\n"}'
```
JPrevost: I am aware I stated I didn't think we needed a client library for this simple GraphQL query... however, I sort of disagree with myself now that I see this.

I had looked at a nice GraphQL library GitHub provides in this experimental work a while back and I feel like it may be worth it. If you have a bit of time maybe take a look at what I did and think about whether you think it might make this more readable / maintainable over time:

https://github.com/MITLibraries/timlight/pull/6/files#diff-c2795793c537f0d56af3be0de530a0adR1

matt-bernhardt: Following up on this - I haven't tried adding the gem, but some of the syntax that I was using here turns out not to be needed. Specifically, the encoding of newlines is superfluous. Further commits on this method are now available.

JPrevost: Let's continue with your initial approach to completion of the work. We can consider a refactor around this in the future. I'll update the related Jira ticket to reflect this.

app/models/search_timdex.rb
```diff
@@ -0,0 +1,35 @@
...
+    @results = {}
+  end
+
+  # Run a search
+  # @param term [string] The string we are searching for
+  # @return [Hash] A Hash with search metadata and an Array of {Result}s
+  def search(term)
+    @query = '{"query":"{search(searchterm: \"' + term + '\", source: \"MIT ArchivesSpace\") {hits records {sourceLink title summary} } }"}'
+    results = HTTP.headers(accept: 'application/json',
+        'Accept-Encoding': 'gzip, deflate, br',
```
JPrevost: As we are not going to switch to the GraphQL gem I want to revisit the flags here. This is not a comment to change anything, it's mostly a note to myself to actually check these settings which I mostly skimmed thinking we might rip this all out before it landed anyway which is no longer the plan :)

app/models/normalize_timdex.rb
```diff
@@ -20,10 +20,14 @@ def view_more(q)
   def extract_results(results, norm)
     return unless results['data']['search']['records']
     results['data']['search']['records'].first(5).each do |record|
-      result = Result.new(record['title'], record['sourceLink'])
+      result_title = record['title'] + ' (' + record['identifier'].gsub('MIT:archivespace:','') + ')'
```
JPrevost: Take a look at this option and consider if it may be clearer than `gsub` here:
https://ruby-doc.org/core-2.7.0/String.html#method-i-delete_prefix

(ruby is all about syntactic sugar and you'll either really appreciate this or think it's the worst lol)

README.md
```diff
@@ -65,6 +65,8 @@ custom hint metadata.
   If exceeded, 'et al' will be appended after this number.
 - `RECAPTCHA_SITE_KEY`
 - `RECAPTCHA_SECRET_KEY`
+- `TIMDEX_URL`: The GraphQL endpoint for Timdex/ArchiveSpace
+- `TIMDEX_SEARCH_URI`: the base search URL for the Timdex/ArchiveSpace application
```
JPrevost: We probably want to call this `ASPACE_SEARCH_URI` or something similar as it may end up really confusing if we use TIMDEX for other content later on.

matt-bernhardt: Yeah, good call. I'm changing this in a forthcoming commit.

app/models/search_timdex.rb
```diff
@@ -0,0 +1,35 @@
...
+#
+# == See Also
+# - https://timdex.mit.edu/playground
+class SearchTimdex
+  attr_reader :results
+
+  TIMDEX_URL = ENV['TIMDEX_URL'].freeze
+
+  def initialize
+    @timdex_http = HTTP.persistent(TIMDEX_URL)
```
JPrevost: I think this persistent connection setup here isn't used afterwards? We should use it though so moving the headers config up here and then calling the post using this setup object would be 🌈 

I'm imagining something like (which drops some headers I don't think we need and updates origin:
```
    @timdex_http = HTTP.persistent(TIMDEX_URL)
    @timdex_http.headers(accept: 'application/json',
      'Accept-Encoding': 'gzip, deflate, br',
      'Content-Type': 'application/json',
      'Origin': 'https://lib.mit.edu'
    )
```

app/models/search_timdex.rb
```diff
@@ -0,0 +1,35 @@
...
+    @results = {}
+  end
+
+  # Run a search
+  # @param term [string] The string we are searching for
+  # @return [Hash] A Hash with search metadata and an Array of {Result}s
+  def search(term)
+    @query = '{"query":"{search(searchterm: \"' + term + '\", source: \"MIT ArchivesSpace\") {hits records {sourceLink title identifier publicationDate physicalDescription summary contributors { value } } } }"}'
+    results = HTTP.headers(accept: 'application/json',
+        'Accept-Encoding': 'gzip, deflate, br',
```
JPrevost: I think it may be better to move these headers to the `initialize` method where you setup a persistent HTTP connection and then call it down here like:

```ruby
results = @timdex_http.post(TIMDEX_URL, :body => @query)
```

app/models/search_timdex.rb
```diff
@@ -0,0 +1,35 @@
...
+        'Content-Type': 'application/json',
+        'Connection': 'keep-alive',
```
JPrevost: If you use the persistent settings you setup in `initialize`, I believe you can drop explicitly stating to keep alive.

app/models/search_timdex.rb
```diff
@@ -0,0 +1,35 @@
...
+        'DNT': '1',
```
JPrevost: I think we can drop DNT as it doesn't make a lot of sense in terms of API usage.

app/models/search_timdex.rb
```diff
@@ -0,0 +1,35 @@
...
+        'Origin': 'https://timdex.mit.edu'
```
JPrevost: I think we'd be better off with other not setting origin or setting it to `https://lib.mit.edu`

test/models/normalize_timdex_test.rb
```diff
@@ -0,0 +1,54 @@
...
+class NormalizeTimdexTest < ActiveSupport::TestCase
+  def timdex_records
+    VCR.use_cassette('popcorn timdex',
+                     allow_playback_repeats: true) do
+      raw_query = SearchTimdex.new.search('popcorn')
+      NormalizeTimdex.new.to_result(raw_query, 'popcorn')
+    end
+  end
+
+  def aspace_records
```
JPrevost: Nice use of functions to prevent having to do this setup in each test. I often forget to do this 🍻

app/assets/stylesheets/_results.scss
```diff
@@ -84,6 +89,10 @@
       content: " | ";
       color: $gray-l2;
     }
+
+    &:first-child:before {
```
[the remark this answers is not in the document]
matt-bernhardt: Line 93 is a way of selectively undoing the rule on line 89. Line 89 creates a pipe character as a visual separator between the type and year fields in the Books panel, but for the new Archives panel that type field isn't present - meaning that for ASpace results, there's a line that gets rendered as:
```
| Date Range: 1905-1974
```

The rule on line 93 prevents this, and suppresses the pipe if the year element is the first child of its parent.

app/models/normalize_timdex.rb
```diff
@@ -0,0 +1,54 @@
...
+    return title
+  end
+
+  def creators(contributors)
+    contributors&.map do |creator|
+      [creator['value'], "&field[]=creators_text&q[]=" << URI.encode_www_form_component(creator['value'])]
+    end
+  end
+
+  def transfer_field(value)
```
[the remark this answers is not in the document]
matt-bernhardt: Both the `transfer_field` and `transfer_first` are meant to be methods that populate the `Result` object from what comes back in the `record` hash. The immediate motiviation was getting flagged for a high degree of cognitive complexity in the `extract_results` method, so maybe these should be `extract_string` and `extract_first_from_list` or something?

test/models/normalize_timdex_test.rb
matt-bernhardt: Thanks for providing the timdex_records function that I could copy - I'm not sure I'd have caught this otherwise. 🍻

app/models/normalize_timdex.rb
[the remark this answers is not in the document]
JPrevost: yeah. My brain didn't follow the `transfer` naming.

README.md
```diff
@@ -65,6 +65,8 @@ custom hint metadata.
   If exceeded, 'et al' will be appended after this number.
 - `RECAPTCHA_SITE_KEY`
 - `RECAPTCHA_SECRET_KEY`
+- `TIMDEX_URL`: The GraphQL endpoint for Timdex/ArchiveSpace
+- `ASPACE_SEARCH_URI`: the base search URL for the Timdex/ArchiveSpace application
```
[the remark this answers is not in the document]
matt-bernhardt: Sure thing - I've added a commit to this branch to modify `app.json` and added values in Heroku.

app/views/search/_result.html.erb
```diff
@@ -41,9 +43,21 @@
...
+    <% if result.physical_description %>
+      <p class="result-physical">
+        Physical Description: <%= result.truncated_physical %>
+      </p>
+    <% end %>
+
     <% if result.blurb %>
       <p class="result-blurb">
-        <%= safe_output(result.blurb) %><br />
+        <%= safe_output(result.truncated_blurb) %><br />
```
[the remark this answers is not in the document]
matt-bernhardt: Yep - the blurb field is populated in `normalize_google`, but not in `normalize_eds*`. The population, though appears to be pretty short - at least if I go by results like https://lib.mit.edu/search/bento?q=chomsky. Google uses an `html_snippet` field that appears to already be pretty short, so I'm not sure this change will actually impact Google results either.

app/models/search_timdex.rb
matt-bernhardt: This change has been implemented. I ended up needing a single statement here rather than two, as you'll see in the diff.

JPrevost:
@matt-bernhardt I've reviewed all of the code on this to this point. 👍 

You can now squash and rebase so we have a simpler branch to maintain while we await stakeholder sign off on the path forward on this.

JPrevost:
Confirming code to this point is reviewed and 👍 . We're still waiting for stakeholder sign off of a launch plan before merging.

PR #634, opened 2020-01-28
Dockerfile

JPrevost:
## Status
**READY**

#### What does this PR do?

- Updates ruby to 2.6.5
- Adds optional local development with Docker option for those that don't
  want to, or can't, setup a stable ruby environment locally

This is not intended to be a container used in production and should only be used in development if you can't run specific version of ruby easily (for instance, on Windows)

Follow the instructions in the readme about building, running the container, and booting to bash to run tests.

NO

NO


JPrevost:
@jazairi typo fixed in the latest commit. Thanks for catching that!

jazairi:
@JPrevost No problem. I assume this is why y'all hired an English major. :)

matt-bernhardt:
I've been able to run the tests on this branch, and get it running (via a few kludges) on my laptop. There's nothing wrong with this work, but I need to streamline how things like Docker are running for me.

I suspect (but haven't confirmed) that the errors I was getting on the tests and running app have to do with case insensitivity on the Windows side of things - but this is a hunch.

PR #635, opened 2020-01-28
Update dependencies

jazairi:
## Status
READY

#### What does this PR do?
Updates the following dependencies: webmock, skylight, http.

NO

YES


jazairi:
Reducing test coverage to 3.318% while updating dependencies is what I like to call a learning opportunity.

JPrevost:
@jazairi I think the issue is the EDS gem. I'd try to revert that back to the old version. We can chat about how to update that one but I've been ignoring it for quite a while because it was going to be moderately annoying.

jazairi:
@JPrevost Looks like mocha and staccato might also be problematic. I'll try rolling those back.

PR #646, opened 2020-02-28
Update dependencies

jazairi:
## Status
IN DEVELOPMENT

#### What does this PR do?
Updates dependencies.

#### Todo:
- [ ] Tests
- [ ] Documentation
- [ ] Stakeholder approval

NO

YES


jazairi:
@JPrevost I still need to figure out what in ebsco-eds and mocha are causing the build to break. (Not worried about staccato, since it needs to go away anyway.) Maybe we could take a look on Monday? Or we could merge this one and take care of those two in separate PR(s).

JPrevost:
@jazairi yeah we can do those separately. I'm 99% sure ebsco is just in need of some cassette updates (which gets complicated... we should chat about how to determine if that is really the problem and how to fix it)

This seems good to me. :shipit:

PR #664, opened 2020-03-30
Upgrade to Rails 5.2

jazairi:
## Status
**READY**

#### What does this PR do?
Upgrades Rails to 5.2.4. Among other things, this adds support for bootsnap, Active Storage, and a built-in redis store. I haven't configured redis yet, but I can if it would be useful. Likewise, I configured Active Storage, but we might not want to bother with it.

Release notes: https://guides.rubyonrails.org/5_2_release_notes.html

- https://mitlibraries.atlassian.net/browse/DI-

#### Todo:
- [x] Tests
- [ ] Documentation
- [x] Stakeholder approval

NO

YES


jazairi:
So, remaining questions I have on this:
* Redis or no?
* ActiveStorage no?

I'm guessing we don't need ActiveStorage, but curious about our thoughts on Redis.

JPrevost:
No ActiveStore or Redis for now. We do use Rails caching with a non-redis backend right now. We'll have to check in on that on Heroku to see how it's working. I suspect based on my quick look we're fine.

Gemfile
```diff
@@ -16,7 +17,7 @@ gem 'lograge'
 gem 'mitlibraries-theme'
 gem 'nokogiri'
 gem 'puma'
-gem 'rails', '5.1.7'
+gem 'rails', '5.2.4'
```
JPrevost: I think there's a `5.2.4.2` out that patched a recent security vulnerability we should bump to while we're at it

PR #711, opened 2020-07-13
Handles record not found errors from EDS gem

JPrevost:
This better handles a case that changed with the update of the EDS gem and treats a record not found as a 404 instead of an exception.

(coming soon)

- https://mitlibraries.atlassian.net/browse/ENGX-44

NO

NO


JPrevost:
This was merged prior to review as it was causing more issues than I expected in our logging quotas. I'll still request a review but need to get this in prod.

PR #714, opened 2020-07-14
Don't throw exceptions on no record found

JPrevost:
## Status
**READY**

#### What does this PR do?

For some reason we were sending exceptions on every invalid record request. We also sent 404 status but throwing an exception was not appropriate here so this removes that.

Locally, check out the master branch and request an invalid record such as:

localhost:3000/record/cat00916a/mit.fake

Note the exceptions in the logs.

Check out this branch and request the same record. You should see no exceptions logged.

- https://mitlibraries.atlassian.net/browse/ENGX-44
NO

NO

### DIFF
--- a/app/controllers/record_controller.rb
+++ b/app/controllers/record_controller.rb
@@ -6,6 +6,12 @@ class RecordController < ApplicationController
   class DbLimitReached < StandardError; end
   class UnknownEdsError < StandardError; end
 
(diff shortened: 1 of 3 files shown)


JPrevost:
1. yes
2. yes, but oddly enough I think the old way returned a 404 and threw an exception

PR #729, opened 2020-08-24
Request digital copy button

JPrevost:
#### What does this PR do?

Enables `Request digital copy` button for appropriate Aleph records.

We're using the same logic as the button in Aleph.
https://docs.google.com/document/d/15Z6gAu4E0784QwYEXH-6kcO8TqHVTfoe_i-dzusGPt8/edit#heading=h.aldraxogy28l

UX will be confirming this so likely just focus on the code and not the "it works as expected" part.

That said, you can use the PR build to test various types of records to ensure the Request Digital button shows up and works as expected... ask in Slack for sample records if you want to see the records I had in developing this. 

- https://mitlibraries.atlassian.net/browse/DI-829
- https://mitlibraries.atlassian.net/browse/DI-830

#### Todo:
- [ ] Tests
- [x] Documentation
- [x] Stakeholder approval

NO

NO

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,10 +87,14 @@ custom hint metadata.
 - `LOG_LEVEL`: set log level for development, default is `:debug`
 - `RESULTS_PER_BOX`: defaults to 3
 - `SENTRY_DSN`: logs exceptions to Sentry
+- `SFX_HOST`: URL to SFX Host. Defaults to production
+- `SFX_ALEPH_HOST`: domain (not full URL) of Aleph server SFX uses to do MARC
+  lookups. Defaults to production.
 - `DISABLE_SCANS`: if set will disable scans for local aleph items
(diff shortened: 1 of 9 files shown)


JPrevost:
NOTE: we cannot merge this until we move away from the test servers being hard coded in the new button. I'll update that to be config as well as add unit tests but wanted to get this in the hands of stakeholders as this is time sensitive.

JPrevost:
I don't think we should address the Code Climate warning. I agree with what is it concerned with, but solving this in a not more confusing way is outside the scope of this work. I'll flag it as such in Code Climate (i.e. yes we should address it but no we won't for this) once others have had a chance to confirm that perspective.

matt-bernhardt:
I'm still at :shipit: with the last two commits.

PR #741, opened 2020-09-17
Hold button tweaks

JPrevost:
## Status
**READY**

#### What does this PR do?

Adjust logic for Hold buttons based on stakeholder feedback.

- Ensures Hold buttons are not displayed for renovation affected items even if WorldCatinator feature is disabled
- Allows Holds on any item status if Recalls are disabled

There are example records in the linked ticket, but mostly we need to check with stakeholders for edge case detection to know for sure if this is 🌈 

- https://mitlibraries.atlassian.net/browse/DI-917

NO

NO

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,8 @@ custom hint metadata.
 ## Optional Environment Variables
 
 - `ALEPH_HINT_SOURCE`: HTTP GET accessible marcxml source for Hints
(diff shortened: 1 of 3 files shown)


jazairi (approved):
Thanks for the detailed commit messages! I think this approach makes sense. If it works for the stakeholders, then :shipit:.

jazairi (approved):
This looks good to go. I don't know Aleph well enough to confirm the query params, so I'll leave that to the stakeholders.

PR #753, opened 2020-10-02
All request digital for available reserve items

JPrevost:
- Status 24 items (high use, usually on resevre) not eligible for Request Item, but need to be eligible for Request Digital Copy

https://mitlibraries.atlassian.net/browse/DI-932

- Also removes the note to ask at desk for reserve items

NO

NO

### DIFF
--- a/app/models/concerns/processing_labeler.rb
+++ b/app/models/concerns/processing_labeler.rb
@@ -89,8 +89,7 @@ module ProcessingLabeler
   # the next action they should take or the amount of delay they should
   # expect. This string is based on the item status in Aleph, but we edit it
   # for readability.
-  def processing_label(status, z30_item_process, reserve)
-    return 'On reserve (see desk)' if reserve
+  def processing_label(status, z30_item_process)
     return 'Checked out' if checked_out?(status)
     return 'In Transit/Sublibrary; Requested' if in_transit?(status)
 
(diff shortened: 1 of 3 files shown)


app/models/concerns/processing_labeler.rb
```diff
@@ -90,7 +90,6 @@ def in_transit?(status)
   # expect. This string is based on the item status in Aleph, but we edit it
   # for readability.
   def processing_label(status, z30_item_process, reserve)
-    return 'On reserve (see desk)' if reserve
```
matt-bernhardt, line 93: Are we using the reserve parameter anywhere else in this method (or related code)? Or can we remove it from the method and the calling code entirely?

JPrevost: Good question.

Yes we still use it for some of the general button logic (some of which is deactivated as this time, but will come back one day with some ENV changes when needed).

JPrevost: Oh wait, I think I misunderstood your question. We indeed don't use it in this method and can likely clean this up. Let me check (my response was based on we still care about reserves elsewhere but you clearly asked about this method and I answered wrongly... sorry). Great catch! Will check and likely clean this up.

JPrevost: @matt-bernhardt I've made the suggested change. Thanks.

matt-bernhardt: Yep - that's what I was thinking of. Thanks! Thumbing next

matt-bernhardt (approved):
This looks good - :shipit:

PR #831, opened 2021-04-07
Update dependencies

jazairi:

Several dependencies needed updating, including the theme gem.

#### Relevant ticket(s):

N/A

Updates the following dependencies:
* mitlibraries-theme
* simplecov
* binding_of_caller
* bootsnap
* barnes
* rack-attack
* ebsco-eds

Sprockets 4 requires a manifest.js file, which I've added to
`app/assets/config/`.

NO

YES


JPrevost (changes requested):
@jazairi can you confirm if the manifest.js needs to be updated? I think we're still on sprockets on this app so that may be detritus from something else?

JPrevost (approved):
Never mind, read your notes. :shipit:

PR #832, opened 2021-04-07
Update pull request template

jazairi:

We've changed the way we structure and commit messages in recent
projects. Now that we use commit message templates, the current Bento
PR template asks for much of the information that is now included in
our commit message templates.

#### Relevant ticket(s):

None.

This updates the PR template to match what we currently use in Thing.

None.

NO

NO


JPrevost (approved):
This has the dep updates from the other PR as well so if it's easier just merge this and close the other PR. Up to you.

PR #842, opened 2021-05-03
IMP-1953 Add normalization models for Primo searches

jazairi:

The app can currently search Primo and return raw results, but it cannot
normalize those results and map them into our Result model.

#### Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/IMP-1953

This adds the relevant normalization models for Primo to map Primo search
results (both Alma and CDI) into our Result model to render in the result
partial. Primo searches can now be seen in the search_boxed view using
the 'alma' and 'cdi' targets.

* Ex Libris has minimal useful documentation on the Primo Search API, so
this code will likely need to be refined over time.
* In an effort to keep things as simple as possible, this changeset
largely mirrors the logic for EDS normalization and minimizes changes
to the Result model and corresponding views. Because of this, some of the
code is a little more convoluted than if we were starting from scratch.
* We are constructing thumbnail URLs to Syndetics using the ISBN. A
PNX record can have multiple ISBNs, so we are taking the first entry
for now.
* All PNX fields are returned as arrays, so it is unclear which are
supposed to be multi-value and which are single-value. We are making some
assumptions to this effect.

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/app/models/normalize_primo_articles.rb
+++ b/app/models/normalize_primo_articles.rb
@@ -1,0 +1,27 @@
+# Tranforms results from {SearchPrimo} into normalized {Result}s
+# for metadata that is relevant for article type results
+class NormalizePrimoArticles
+  def initialize(record)
+    @record = record
+  end
+
+  def article_metadata(result)
+    result.citation = numbering
+    result.in = journal_title
+    result
+  end
+
+  def journal_title
+    return unless @record['pnx']['addata']['jtitle']
+    @record['pnx']['addata']['jtitle'].join('')
+  end
+
+  def numbering
+    return unless @record['pnx']['addata']['volume']
+    if @record['pnx']['addata']['issue'].present?
+      "volume #{@record['pnx']['addata']['volume'].join('')} issue #{@record['pnx']['addata']['issue'].join('')}"
+    else
+      "volume #{@record['pnx']['addata']['volume'].join('')}"
+    end 
+  end
+end
--- a/app/models/normalize_primo_books.rb
+++ b/app/models/normalize_primo_books.rb
@@ -1,0 +1,49 @@
+# Tranforms results from {SearchPrimo} into normalized {Result}s
+# for metadata that is relevant for book type results
+class NormalizePrimoBooks
+  def initialize(record)
+    @record = record
+  end
+
+  def book_metadata(result)
+    result.thumbnail = thumbnail
+    result.publisher = publisher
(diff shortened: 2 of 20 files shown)


jazairi:
I'll leave the Code Climate issues open pending code review. One is due to excessive cognitive complexity, and the others are for not being DRY. The lack of DRYness is by design, but I'm happy to reconsider if the reviewer feels it should change.

app/models/normalize_primo_articles.rb
```diff
@@ -0,0 +1,23 @@
...
+    result
+  end
+
+  def journal_title
+    return unless @record['pnx']['addata']['jtitle']
+    @record['pnx']['addata']['jtitle'].join('')
+  end
+
+  def numbering
+    return unless @record['pnx']['addata']['volume'] && @record['pnx']['addata']['issue']
```
JPrevost, line 20: I don't have an example record, but I assume some journals have volumes but not issues.

I'm not opposed to leaving this as is with an additional ticket to return to it though.

app/models/normalize_primo_books.rb
```diff
@@ -0,0 +1,49 @@
...
+  def book_metadata(result)
+    result.thumbnail = thumbnail
+    result.publisher = publisher
+    result.location = location
+    result.subjects = subjects
+    result
+  end
+
+  def thumbnail
+    return unless @record['pnx']['addata']['isbn']
```
JPrevost, line 17: We may want to consider using issn on the articles normalizer to add thumbnails as well. (not asking for a change to this PR, just noting I think syndetics can also accept issn and may be worth asking UX if they want that)

JPrevost (approved):
👍🏻 

I won't fixed a few code climate and accepted a couple to return to later.

My comments are non-blocking on this work. I do think we should open new tickets for the volume / issue question and the syndetics for journals question.

app/models/normalize_primo_articles.rb
jazairi: Oh, that's a good point. I'll refactor to account for that.

app/models/normalize_primo_books.rb
jazairi: Good call. I'll make a note in the QA ticket.

JPrevost (approved):

jazairi:
As per our usual practice, stakeholder review will occur in staging.

PR #883, opened 2021-06-28
IMP-2047 Handle irregular Alma openurl links

jazairi:

The app currently expects all Alma openurls to use the 'na06'
Ex Libris server. This could cause problems if we receive records that
use a different server for their openurls.

#### Relevant ticket(s):

https://mitlibraries.atlassian.net/browse/IMP-2047

* Logs all records that use a server other than na06 for their Alma
openurl.
* Returns the regular Alma openurl in the record instead of constructing
a Primo openurl for records that don't use the na06 server.

Testing this logic required an additional customization to the
missing_fields_primo_articles cassette. This is already a customized
cassette, and the additional customization has been noted in the
normalize_primo_articles model test.

#### Developer

#### Code Reviewer

NO

NO


jazairi:
I'm marking stakeholder approval for this one because the change doesn't affect the UI. (Technically it does, but I'm not sure how to demonstrate for UX since we don't know of any records with irregular Alma openurls.)

app/models/normalize_primo_articles.rb
```diff
@@ -43,10 +43,24 @@ def chapter_numbering
...
+    openurl_server = ENV['ALMA_OPENURL'][8,4]
+    record_openurl_server = @record['delivery']['almaOpenurl'][8,4]
+    if openurl_server == record_openurl_server
+      primo_openurl = ['https://mit.primo.exlibrisgroup.com/discovery/openurl?institution=',
+                       ENV['EXL_INST_ID'], '&vid=', ENV['PRIMO_VID'],'&'].join('')
+      link_resolver_url = @record['delivery']['almaOpenurl'].gsub(ENV['ALMA_OPENURL'],
+                                                                  primo_openurl)
+    else
+      Rails.logger.warn 'Alma openurl server mismatch. Expected '
+                         + openurl_server + ', but received ' +
```
JPrevost, line 59: This is weird, but if you move this first `+` from this line to the line above the logging works as expected but like this is only prints the first line. 🤷🏻

JPrevost (changes requested):

app/models/normalize_primo_articles.rb
jazairi: Thanks, that's resolved in the latest commit. I also updated the books version of the method to use the Primo URL env, because I noticed while looking at this that it was hardcoded.

JPrevost (approved):

PR #904, opened 2021-08-16
IMP-2283 Remove target validation from Search Controller

jazairi:

The Search Controller contains some confusing logic with regard to
validating the target param. The controller checks the target param
against an array of 'valid' targets; if the target is not
considered valid, the controller returns a redirect to the root URL.

This is problematic for a few reasons:

1. It's not very maintainable. We will have to modify the hardcoded
array of valid targets whenever the value of the param, which is stored
in ENV, changes. We've already encountered this with faked param values
in our test suite.
2. It's not very usable. If a user enters a bad param, redirecting
them to the root URL doesn't tell them anything useful about what
went wrong.
3. It's not very intuitive. We already have a method that checks for
which type of search to run based on the target provided. This seems
like the logical place to check for bad targets, rather than three
separate places as we do now.
4. It's not very necessary. This is what 404 errors are for!

#### Relevant ticket(s):

https://mitlibraries.atlassian.net/browse/IMP-2283

This removes the aforementioned target validation logic from the
Search Controller, and instead returns a 404 error (and renders our
custom 404 template) when it gets a bad target param. Note that this
only occurs from the search_boxed view. The bento view continues to
ignore additional params, target or otherwise.

A custom error is used here to avoid a DoubleRenderError.

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -1,6 +1,10 @@
 class SearchController < ApplicationController
   before_action :validate_q!, only: %i[bento search search_boxed]
 
+  class NoSuchTargetError < StandardError; end
+
+  rescue_from NoSuchTargetError, with: :no_such_target
+
   def index; end
 
   def bento; end
@@ -8,7 +12,6 @@ class SearchController < ApplicationController
   def search_boxed
     @per_box = (ENV['RESULTS_PER_BOX'] || 5).to_i
     @results = search_results(1, @per_box)
-    return redirect_to root_url unless @results
     render layout: false
   end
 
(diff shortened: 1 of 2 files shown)


jazairi:
I'm not concerned about the codeclimate issue, primarily because it will resolve itself once we remove EDS search functionality, which is next up for this app.

The test coverage drop is confusing to me, and I could use some help diagnosing it. Coveralls shows a drop in the controllers directory, but none of the individual controller files have reduced coverage. Am I missing something?

matt-bernhardt (approved):
This all looks good to me - I like the posture of raising an error in the "else" clause of the `search_target` method, rather than having the default behavior to run a specific search. It took me a while to wrap my head around raising an error, only to rescue from it and then render a partial, but I think I follow it now.

The CodeClimate issue isn't a concern to me. 5 is a pretty arbitrary limit for cognitive complexity in this case, and you're right that it will drop again pretty soon.

As for the test coverage issue, I think what's happening is that the whole controller now has fewer testable lines - so even though it remains 100% tested, the overall percentage of covered lines in all controllers decreases, because each line is now "worth more". So those few untested lines go from, say, 5 / 100 (or 5%) to 5 / 99 (or 5.05%).

:shipit:

PR #951, opened 2021-12-16
ES-572 Primo openurl bugfix

jazairi:
Why these changes are being introduced:

Some Primo openurls do not resolve correctly. The Alma/Primo team
suspects that the cause is the ctx params in the URL.

Relevant ticket(s):

https://mitlibraries.atlassian.net/browse/ES-572

How this addresses that need:

This removes the ctx params from the Primo openurls.

Side effects of this change:

I have some reservations about how much we're manipulating URL
strings here. If this fixes the problem, we should have a
conversation about this in code review.

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/app/models/normalize_primo_articles.rb
+++ b/app/models/normalize_primo_articles.rb
@@ -37,10 +37,6 @@ class NormalizePrimoArticles
     "#{@record['pnx']['addata']['date'].join('')}, pp. #{@record['pnx']['addata']['pages'].join('')}"
   end
 
-  # Here we are converting the Alma link resolver URL provided by the Primo 
-  # Search API to redirect to the Primo UI. This is done for UX purposes, 
-  # as the regular Alma link resolver URLs redirect to a plaintext 
-  # disambiguation page.
   def openurl
     return unless @record['delivery']['almaOpenurl']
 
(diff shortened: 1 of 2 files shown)


jazairi:
Adam Shire and Darcy confirmed that this solves the problem, so it's ready for review and deployment.

app/models/normalize_primo_articles.rb
```diff
@@ -61,4 +55,21 @@ def openurl
...
+  # disambiguation page.
+  def construct_primo_openurl
+    return unless @record['delivery']['almaOpenurl']
+
+    primo_openurl_base = [ENV['MIT_PRIMO_URL'], '/discovery/openurl?institution=', ENV['EXL_INST_ID'], '&vid=',
+                          ENV['PRIMO_VID'], '&'].join('')
+    primo_openurl = @record['delivery']['almaOpenurl'].gsub(ENV['ALMA_OPENURL'], primo_openurl_base)
+
+    # The ctx params appear to break Primo openurls, so we need to remove them.
+    params = Rack::Utils.parse_nested_query(primo_openurl)
```
JPrevost, line 71: Can we add tests to confirm the ctx_xyz parameters are being scrubbed?

JPrevost: Oh wait, the existing test does confirm we scrub that I think...nevermind.

app/models/normalize_primo_articles.rb
```diff
@@ -61,4 +55,21 @@ def openurl
...
+    filtered = params.delete_if { |key, value| key.include?('ctx') }
```
JPrevost, line 72: It feels like there is an edge case that I just want to make sure we are comfortable with. I think we are, but want to note it. A parameter of `rft.ctxblahblahblah` would be scrubbed. I think that is unlikely enough to not care about, but wanted to be sure you agree this implementation is good enough and a `start_with?` instead of `include?` would not be better based on what we know.

jazairi: That's a good point. I think `start_with?` is a good idea for now.

jazairi: Hm...actually, I'm having second thoughts about this. I think we would want to scrub, for example, something like this: `url_ctx_fmt=info:ofi/fmt:kev:mtx:ctx`

JPrevost: Yeah not knowing what Primo can actually handle is annoying. I'm comfortable with however we handle this as regardless of what we do we'll likely

JPrevost (approved):

PR #1009, opened 2022-04-06
Updates local rubocop implementation

matt-bernhardt:
#### Why are these changes being introduced:

* We haven't kept up to date with our use of rubocop locally, leading it
  to need some updates to get to the point where `rubocop -a` can be
  run successfully.

#### Relevant ticket(s):

* n/a

#### How does this address that need:

* This adds the `rubocop-rails` gem, which was split out from the main
  `rubocop` gem at some point in the past.
* The format of `.rubocop.yml` is also updated to match what is in our
  other Rails apps

Taken together, this now means that we can run rubocop locally, and in
a manner similar to our other apps.

#### Document any side effects to this change:

* Re-alphabetized the Gemfile in one instance

#### Developer

- [ ] All new ENV is documented in README

#### Code Reviewer

YES | NO

YES | NO


JPrevost:
@matt-bernhardt would you mind incorporating #1010 with this PR?

matt-bernhardt:
Okay, @JPrevost or @jazairi - I think this is now ready to ask for code review. I've tried to structure the commits in a way to make them easy to review:

1. First commit fixes the local rubocop implementation, allowing us to run things in a way that should be similar to what CodeClimate does (not exactly, though - see the final note)
2. Second commit makes all changes recommended by `rubocop -a` - I ran the test suite afterwards, but didn't really look at the details here. I've no reason right now to not trust that step.
3. At this point I ran `rubocop -A` and looked at the changeset. There's a lot of changes, some of which attracted my attention - so I tried to group them. This commit is entirely focused on default arguments for `split()` and `join()` calls, which seem okay to me. There's a lot of them, so having one commit just for these allowed me to focus on the more substantive changes.
4. This commit groups a number of one-off changes. The commit message lists what changes get made. I don't feel like arguing with a linter on these, and the new syntax seems okay. I kind of like case statements over long if/else changes, for example. The tests still pass, and we've got good coverage in this app.
5. This is the bit that I'm most curious about. In theory I like string interpolation like this, but there was another change suggested which I've decided to ignore. Curiously, while my local rubocop wants to change this line: https://github.com/MITLibraries/bento/blob/main/app/models/search_timdex.rb#L27 - it isn't being flagged by CodeClimate. This is just as well IMO, because I don't agree with the change being suggested.

The change is to change:
```
@query = '{"query":"{search(searchterm: \"' + clean_term(term) + '\", source: \"MIT ArchivesSpace\") {hits records {sourceLink title identifier publicationDate physicalDescription summary contributors { value } } } }"}'
```
into
```
@query = "{\"query\":\"{search(searchterm: \\\"#{clean_term(term)}\\\", source: \\\"MIT ArchivesSpace\\\") {hits records {sourceLink title identifier publicationDate physicalDescription summary contributors { value } } } }\"}"
```
...and that many escaping slashes just looks painful to me. Since it isn't getting flagged by CodeClimate I'm not sure if we're going to hear about this in the future. A better approach to this whole method might be to build a hash and use that to somehow build the graphql query - but I've not seen a great way to do that yet. The GraphQL syntax is weird, so building a string may be our best option - or maybe we just leave this as-is and tackle the question later.

Interestingly, one of the changes that gets made locally is now causing a line length flag - I'm inclined to ignore the line length issue, because the code feels more readable now rather than breaking that message into what ends up being short snippets due to indenting. I'm open to counter opinions on this, though.

---
Anyway - I think this is ready to merge now. Let me know what you think?

JPrevost (approved):
Thanks for breaking down the commits and explaining them clearly. It made this much easier to review than it would have been otherwise.

:shipit:

PR #1137, opened 2023-01-25
Fix source param for TIMDEX V2 searches

jazairi:

In V2, all facets have been renamed so it is clearer that they are facets. This means that V2 searches in Bento will fail, because the search string uses `source` instead of `sourceFacet`.

More information in this commit:
https://github.com/MITLibraries/timdex/commit/3f2801b6644eca83ac84ae678864f07b4f9c75e9

#### Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/TIMX-170

This adds a timdex_v2 feature to Flipflop, which is used to alter the search string to correspond with the target TIMDEX version. When this feature is enabled, `source` becomes `sourceFacet`.

* In order to make both V1 and V2 searches succeed, I had to run each set of tests separately as we do in TIMDEX. Unfortunately, this means reverting from the shared Ruby CI to a custom CI, at least until we shut off V1.
* ClimateControl has been added to the test group to accommodate testing.

#### Developer

#### Code Reviewer

NO

YES


jazairi:
The missing lcov file makes sense to me, but I'm not sure how to fix it. I assume it would be created on merge?

JPrevost:
@jazairi we standardize to `coverage.lcov` to allow the shared workflows to work so you'd need to update the bespoke ci to use `coverage.lcov` instead of `bento.lcov` I believe.

https://github.com/MITLibraries/bento/blob/main/test/test_helper.rb#L4

JPrevost:
I think we should remove the feature flag and only merge this when we are ready to move to v2 in production (i.e. don't support both `source` and `sourceFacet` and instead just merge the `sourceFacet` work right before we launch v2 in prod. That would allow for all of the CI to remain unchanged. If that makes sense to you please go ahead and rework this to only support `source` but leave it as a draft PR so we don't accidentally merge it :)

PR #1141, opened 2023-01-31
Update Bento to use TIMDEX V2 fields

jazairi:

Bento currently uses deprecated fields in its TIMDEX searches:

* `identifier`
* `publicationDate`
* `source`

#### Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/TIMX-174

This makes the following V1 to V2 replacements:

* `identifier` with `identifiers` (`kind` == 'Collection Identifer')
* `publicationDate` with `dates` (there is some nuance to this; see the changes to the NormalizeTimdex model for details)
* `source` with `sourceFacet`

BEFORE MERGING:
* We will need to wait for prod launch, as V2 endpoints currently have a different URL.
* Cassettes will need to be regenerated with correct URL.

* Some cassettes have been regenerated to use the new search params.
* Tests that anticipate results from non-ASpace sources have been. removed. These tests would fail if we were to regenerate cassettes in the current codebase, since the ASpace source is hardcoded in the search string.

#### Developer

- [ ] All new ENV is documented in README

#### Code Reviewer

NO

NO


app/models/normalize_timdex.rb
```diff
@@ -33,10 +33,36 @@ def extract_results(results, norm)
...
+    title << extract_collection_id(record['identifiers']) if record['identifiers'].present?
     title
   end
 
+  def extract_dates(dates)
+    # It is unlikely for a record to have more than one creation date, but just in case...
+    relevant_dates = dates.select { |date| date['kind'] == 'creation' }
+
+    # If the record *does* have more than one creation date, it's probably not worth determining which to display.
+    return if relevant_dates.count > 1
```
JPrevost, line 45: I'm waffly on whether displaying no dates or the first date is better. (This is not suggesting a change, just saying that both paths seem equally valid here and thus this one is 👍🏻 )

app/models/normalize_timdex.rb
```diff
@@ -33,10 +33,36 @@ def extract_results(results, norm)
...
+
+    "#{relevant_date['range']['gte']}-#{relevant_date['range']['lte']}"
+  end
+
+  def extract_collection_id(identifiers)
+    relevant_ids = identifiers.map { |id| id['value'] if id['kind'] == 'Collection Identifier' }.compact
+
+    # In the highly unlikely event that there is more than one collection identifier, there's something weird going
+    # on with the record and we should skip it.
+    return if relevant_ids.count > 1
```
JPrevost, line 61: This is kinda funny to me. I agree with it, but I wish I trusted we'd never need this line lol

JPrevost (commented):
This seems good. I'll hold off on my approval until we update it to use `sourceFilter` once that work lands.

app/models/normalize_timdex.rb
jazairi: Yeah, I feel the same way. Of course, it's highly unlikely that this will be an issue, but now that I've said that...

app/models/normalize_timdex.rb
jazairi: Hope for the best, plan for the worst, right? That said, I'm open to removing this and the comparable guard clause in the date method. This is the kind of thing I'd always do when handling Alma data, but I know less about the state of our ASpace data.

jazairi:
@JPrevost Just pushed a commit that's close to prod-ready. As I mentioned in Slack, some of the test assertions had to be rewritten to match the latest data load. We will also want to regenerate cassettes once prod is on V2, as they currently call `https://timdex-api-prod-v2.herokuapp.com`. However, there is no rush on this, as current prod Bento will continue to work with V2 (albeit with slightly weird identifiers in the titles).

I'll leave this as a draft until we launch and the tests are regenerated.

JPrevost (approved):

PR #1157, opened 2023-03-14
Better error handling for dates

jazairi:

We encountered a bug after V2 launch in the normalization process, during which Bento hits an unhandled nil value that breaks the search results. This is because we are assuming every ASpace record has a creation date, when in fact some of them have publication dates instead.

Somewhat related to this is a transmogrifier bug that causes a `kind` to be associated with a date even when there is no date.

#### Relevant ticket(s):

* https://mitlibraries.atlassian.net/browse/TIMX-211
* https://mitlibraries.atlassian.net/browse/TIMX-212

This updates `NormalizeTimdex#extract_dates` to select as relevant dates of either the 'creation' _or_ 'publication' date kind. It also adds a guard clause to return if the `relevant_dates` array is empty (a precaution that would have prevented this error).

I removed a superfluous check for `date['kind'] == 'creation'` in the guard clause in the `extract_dates` method that checks for a date range. At this point in the method, we have already confirmed that the relevant date is either a creation or publication date.

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/app/models/normalize_timdex.rb
+++ b/app/models/normalize_timdex.rb
@@ -38,17 +38,18 @@ class NormalizeTimdex
   end
 
   def extract_dates(dates)
-    # It is unlikely for a record to have more than one creation date, but just in case...
-    relevant_dates = dates.select { |date| date['kind'] == 'creation' }
+    # It is unlikely for a record to have more than one creation or publication date, but just in case...
+    relevant_dates = dates.select { |date| date['kind'] == 'creation' || date['kind'] == 'publication' }
 
-    # If the record *does* have more than one creation date, it's probably not worth determining which to display.
-    return if relevant_dates.count > 1
+    # If the record has no creation or publication date, stop here.
+    return if relevant_dates.empty?
 
+    # If the record *does* have more than one creation/pub date, just take the first one.
     relevant_date = relevant_dates.first
 
-    # We are only concerned with creation dates that are ranges, since we harvest ASpace metadata at the collection
+    # We are only concerned with creation/pub dates that are ranges, since we harvest ASpace metadata at the collection
     # level.
-    return unless relevant_date['kind'] == 'creation' && relevant_date['range'].present?
+    return unless relevant_date['range'].present?
 
     "#{relevant_date['range']['gte']}-#{relevant_date['range']['lte']}"
   end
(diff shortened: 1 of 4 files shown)


app/models/normalize_timdex.rb
```diff
@@ -38,17 +38,20 @@ def extract_title(record)
...
-    relevant_dates = dates.select { |date| date['kind'] == 'creation' }
+    # It is unlikely for a record to have more than one creation or publication date, but just in case...
+    relevant_dates = dates.select { |date| date['kind'] == 'creation' || date['kind'] == 'publication' }
 
-    # If the record *does* have more than one creation date, it's probably not worth determining which to display.
+    # If the record has no creation or publication date, stop here.
+    return if relevant_dates.empty?
+
+    # If the record *does* have more than one creation/pub date, it's probably not worth determining which to display.
     return if relevant_dates.count > 1
```
JPrevost, line 48: I think now that we are checking both kinds of dates here we should remove this logic and just use the first value (or prioritize creation date or publication date if we know which is actually preferred here (which I don't)).

test/models/normalize_timdex_test.rb
```diff
@@ -36,4 +36,32 @@ def aspace_records
...
+                     allow_playback_repeats: true) do
+      raw_query = SearchTimdex.new.search('SOVEREIGN INTIMACY: PRIVATE MEDIA AND THE TRACES OF COLONIAL VIOLENCE')
+      assert_not raw_query['data']['search']['records'].first['dates'].first['kind'] == 'creation'
+      assert_nothing_raised do
+        NormalizeTimdex.new.to_result(raw_query, 'SOVEREIGN INTIMACY: PRIVATE MEDIA AND THE TRACES OF COLONIAL VIOLENCE')
+      end
+    end
+  end
+
+  # This is another regression test. Now that we know that records with no creation date will not error, we also want
```
JPrevost, line 53: I like this additional test. Do you think we might want a test to confirm we don't error but still normalize _a_ date when more than one date exists now that we've decided yolo choose one instead of displaying none?

JPrevost: Good point. Hmmm...

JPrevost: The first record when searching for `Timurid Architecture Research Archive` has two creation date ranges that should work for this.

PR #1251, opened 2023-12-04
Ensure total is an integer before comparison

JPrevost:
https://mitlibraries.atlassian.net/browse/PW-81

It's unclear why this only became a problem recently, so this may not address the underlying problem, but it does seem to fix the immediate concern.

#### Developer

#### Code Reviewer

NO

NO

### DIFF
--- a/app/views/search/_view_all_link.html.erb
+++ b/app/views/search/_view_all_link.html.erb
@@ -1,10 +1,10 @@
 <% if current_page?(:controller => 'search', :action => 'search_boxed') %>
   <div class="results-more">
-  <% if @results['total'] > 0 %>
+  <% if @results['total'].to_i.present? %>
     <%= link_to(
       "View more #{view_more_text} results",
       @results['view_more'],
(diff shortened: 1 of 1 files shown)


app/views/search/_view_all_link.html.erb
```diff
@@ -1,10 +1,10 @@
 <% if current_page?(:controller => 'search', :action => 'search_boxed') %>
   <div class="results-more">
-  <% if @results['total'] > 0 %>
+  <% if @results['total'].to_i.present? %>
```
jazairi, line 3: I'm unclear on what would happen here if `@results['total']` can't be cast into an integer for some reason, but that seems like a highly unlikely scenario, and Sentry will let us know if it happens anyway.

jazairi (approved):
This seems like a reasonable patch. I'm not sure why we were concerned with the value of `@results['total']` in that `if` clause since we were conditionally hiding the button anyway, so checking for the presence of that element rather than its value feels like an improvement.

app/views/search/_view_all_link.html.erb
JPrevost: Yeah. I was wary to look too deep as you are correct and this may be problematic... 🤞🏻 

If this leads to more trouble than it solves, we can take more time and fix it right.

PR #1311, opened 2024-05-17
Initial TACOS integration

matt-bernhardt:
This defines a new env variable, TACOS_URL, which stores the endpoint
for the TACOS GraphQL service. When defined, the search results page
will then direct the user agent to send the search string to the TACOS
system, which will eventually respond with any additional information
we can detect about the search string.

For now, the only response will be to reflect the search string back
into the console, along with the originating system.

#### Ticket

- https://mitlibraries.atlassian.net/browse/ENGX-267 (this change to Bento)
- https://mitlibraries.atlassian.net/browse/ENGX-266 (the related TACOS change to CORS behavior)

#### Environment

#### Accessibility
- [x] There are no user-facing UI changes that need a11y testing

#### Stakeholders
- [x] Stakeholder approval is not needed at this time

NO

NO

### Code Reviewer

### DIFF
--- a/app/views/search/_trigger_tacos.html.erb
+++ b/app/views/search/_trigger_tacos.html.erb
@@ -1,0 +1,12 @@
+$.ajax({
+  type: "POST",
+  url: "<%= ENV.fetch('TACOS_URL', '') %>",
+  contentType: 'application/json',
+  data: '{ "query": "{ logSearchEvent( searchTerm: \\"<%= params[:q] %>\\" sourceSystem: \\"bento\\" ) { phrase source } }" }',
+  dataType: 'json'
+}).success(function (msg) {
+  console.log(msg.data.logSearchEvent);
+}).fail(function (xhr, textStatus ) {
+  console.log('Failure looking up search in Tacos:');
+  console.log(xhr);
+});
--- a/app/views/search/_trigger_tacos.html.erb
+++ b/app/views/search/_trigger_tacos.html.erb
@@ -2,11 +2,8 @@ $.ajax({
   type: "POST",
   url: "<%= ENV.fetch('TACOS_URL', '') %>",
   contentType: 'application/json',
-  data: '{ "query": "{ logSearchEvent( searchTerm: \\"<%= params[:q] %>\\" sourceSystem: \\"bento\\" ) { phrase source } }" }',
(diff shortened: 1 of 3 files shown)


app/views/search/_trigger_tacos.html.erb
```diff
@@ -0,0 +1,12 @@
+$.ajax({
```
JPrevost, line 1: General feedback on the approach. None of it is blocking or should be considered reason to change anything without discussion as I'm not looking for changes so much as conversation to help me decide how I feel about some of this work.

- If we keep this as client-side javascript, I think I'd prefer the console logging for non-error states not be on by default for production-like deployments. I'd listen to reasons to keep it on, but it feels like debug data and thus probably doesn't belong in prod to me.
- I know the GraphQL behavior of this is a bit odd, but I think requesting both phrase and source is probably not useful for non-debugging purposes. As we'll just ignore it all, returning just one field seems better for now (I suspect a simplified REST API in addition to the GraphQL endpoint for use by systems that don't care about returned data may be worth considering for us in the future). As we'll hopefully move Hints into Tacos soon we'll have more opportunities to rethink what to request when we get to that.
- We should move all of these `trigger_` events (search, hints, tacos) to a more modern rails approach. That is out of scope for this work, but my gut reaction was "wtf is this mess" and then I realized it was modeled directly on the code I initially wrote for bento which has worked for several years so maybe it isn't as bad as it feels?

JPrevost (commented):
I suspect after some brief discussion I'm move this to an approval, but feel a quick chat will help me with that.

app/views/search/_trigger_tacos.html.erb
JPrevost: After a good discussion, we'll open backlog tickets to refactor all the bento async work to Turbo/Stimulus and during that work consider whether it makes sense to move the GraphQL POST into Rails instead of client-side JS... but for now this works well.

matt-bernhardt:
Okay @JPrevost - I've pushed up some changes based on our conversation yesterday. Let me know what you think?

JPrevost (approved):

JPrevost:
@matt-bernhardt this looks good and the tests actually passed but Actions is having a bad day.

PR #1351, opened 2024-09-09
Replace Hints with Suggested Resources

jazairi:

We've replaced the Hints system built into Bento with the Suggested Resources detector, provided by TACOS API. Since Bento is already logging search terms to TACOS in production, it makes sense to switch to using Suggested Resources.

#### Relevant ticket(s):

* [TCO-66](https://mitlibraries.atlassian.net/browse/TCO-66)

This removes the Hint feature and replaces it with Suggested Resources. We are currently triggering Suggested Resources in the same way as Hints (and the other controllers), but it might make sense to refactor this feature using a tool like Stimulus or Turbo. I've opened (ENGX-283)[https://mitlibraries.atlassian.net/browse/ENGX-283] to track further work on this.

While discussing this feature, we realized that GraphQL returns an array for `suggestedResources` even though that detector will only return a single resource. We are handling this in Bento by just calling `first` on the array, but we should reconsider whether the GraphQL type is returning the correct data structure. The spike ticket [TCO-84](https://mitlibraries.atlassian.net/browse/TCO-84) will further explore this problem space.

* As far as I can tell, we don't really have a notion of hint source anymore, so I haven't tried to replicate the helper methods. This seems most pertinent to HintHelper#data_type, which professes to allow analytics tools to distinguish between hint sources. (It's unclear if this is still used by Matomo.) It also affects the display type, of which we have two for Hints, but that doesn't seem like a significant change.
* Debug and ClimateControl have been added to the Gemfile. The former is what we should be using to debug instead of Pry, and the latter was needed to test various ENV conditions.

#### Developer

- [x] All new ENV has been added to Heroku Pipeline, Staging and Prod *(Added to staging and prod)*

#### Code Reviewer

NO

YES

[TCO-66]: https://mitlibraries.atlassian.net/browse/TCO-66?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[TCO-66]: https://mitlibraries.atlassian.net/browse/TCO-66?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[TCO-84]: https://mitlibraries.atlassian.net/browse/TCO-84?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jazairi:
Merging without review since two engineers have worked on this, as discussed during our demo/retro this morning.

